PR #33, opened 2023-08-23
instrument: refactor, format, and organize code by integration

darccio:
### What does this PR do?

Refactor duplicated code by reusing utility functions like `wrap` from adding `dd:startwrap` and `dd:endwrap`, clean format by removing some blank lines, and organize integrations' code in separate files. Also, all `labstack/echo.v4` code uses `[eE]choV4` as prefix.

### Motivation

`instrument.go` and `uninstrument.go` were getting really big.

### Reviewer's Checklist


ahmed-mez (approved):
🚀

Makefile
```diff
@@ -1,6 +1,6 @@
 .DEFAULT_GOAL := build
 
-.PHONY: build generate vet test clean fmt licenses verify-licenses
+.PHONY: build generate tidy vet test clean fmt licenses verify-licenses
```
darccio, line 3: Question: what is the purpose of `generate` target? I saw it only declared as a phony target, but it doesn't seem to have any use. cc @ahmed-mez @knusbaum @jonbodner

ahmed-mez: It's likely an oversight when we first created the Makefile? Perhaps it was copied from another Go project that has a generate target. As far as Orchestrion is concerned, we don't have any code generation yet. You can get rid of it if you want.

darccio: As it's a minor nit, I'll keep it. Maybe the idea was to run `go:generate` commands? I see one usage of `stringer` in `event.go.`

PR #40, opened 2023-09-07
instrument: report orchestrion config to dd-trace-go

ahmed-mez:
### What does this PR do?

- Start the tracer with the option `tracer.WithOrchestrion`, it achieves the following
  - make the tracer log orchestrion info in startup logs
  - report orchestrion telemetry

This is a draft for now as it depends on the release of dd-trace-go v1.55.0 which hasn't happened yet

### Motivation

- Ease debugging
- Improve telemetry

### Reviewer's Checklist

- [ ] Changed code has unit tests for its functionality


ahmed-mez:
Short answer is no it's not supported. Why would we need to do it? Can't think of a use-case, but it's something we can explore if there is a need.

Since the beginning we assumed that users shouldn't care much about the tracing backend (dd-trace-go, otel), we also made multiple changes in dd-trace-go to improve the auto-instrumentation UX (like having better default attributes in the http contrib, support database/sql without providing a driver explicitly and now with `WithOrchestrion` tracer option). The assumption is that Orchestrion should always use the latest stable dd-trace-go version.

darccio (approved):

PR #48, opened 2023-10-10
instrument/http: add span.kind & component tags to WrapHandlerFunc

darccio:
### What does this PR do?

Adds missing tags `span.kind` and `component` for `net/http` autoinstrumentation using `instrument.WrapHandlerFunc`.

The root cause is that it uses `httptrace.TraceAndServe` under the hood, that assumes these tags are added by the caller (like `httptrace.WrapHandler`).

### Reviewer's Checklist

### DIFF
--- a/_integration-tests/tests/net_http/validation.json
+++ b/_integration-tests/tests/net_http/validation.json
@@ -5,6 +5,10 @@
 	    "name": "http.request",
 	    "service": "net_http",
 	    "resource": "GET /",
-	    "type": "web"
+	    "type": "web",
+		"meta": {
+			"component": "net/http",
+			"span.kind": "server"
+		}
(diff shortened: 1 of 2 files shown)


ahmed-mez (commented):
Any reason why we shouldn't have this logic in dd-trace-go?

darccio:
@ahmed-mez I thought of that, but it seems that `httptrace.TraceAndServe` can be used directly and indirectly through `httptrace.WrapHandler`, that sets these tags before calling `httptrace.TraceAndServe`. This is way I chose to do it in Orchestrion instead.

I agree this should be in `dd-trace-go` but it could break expectations of current users. Maybe it's a good candidate to be included in our v2.

PR #62, opened 2024-01-24
feat: toolexec proxy

Hellzy:
### What does this PR do?

This change adds tool called the toolexec proxy to the repository.
The toolexec proxy's job is to be used with the `-toolexec` go build otption in order to intercept,
inspect and modify go tool commands at compile time.

#### Proxy

The proxy parses go tool commands and returns `Command` interfaces which can be further declined into more
specific commands. Currently two commands types are supported:
- Compile
- Link

Any other command is treated as an uknown command, since we currently only need the two aforementioned commands
for package/code injection.

#### Processors

Processors are command visitors which allow performing a specific action on a command.
Actions can be inspecting the command, modifying its state, bypassing it, etc...
### Motivation

This tool has multiple application cases. It was mostly designed to allow seamless package modification/substitution at compile time in order to facilitate code instrumentation

### Reviewer's Checklist


RomainMuller (commented):
Here's an alternative design I think could be more ergonomic:

```go
// Parse command
cmd := proxy.MustParseCommand(args)

// Register hooks (called immediately if the cmd type matches)
cmd.OnCompile(func (*proxy.CompileCommand){ /* ... */ })
cmd.OnAsm(func (*proxy.AsmCommand){ /* ... */ })
cmd.OnLink(func (*proxy.LinkCommand){ /* ... */ })

// Execute the finalized command
cmd.MustRun()
```

This way, you don't need the empty no-op implementations, and you can more directly target the changes you want to make.

Hellzy:
> Here's an alternative design I think could be more ergonomic:
> 
> ```go
> // Parse command
> cmd := proxy.MustParseCommand(args)
> 
> // Register hooks (called immediately if the cmd type matches)
> cmd.OnCompile(func (*proxy.CompileCommand){ /* ... */ })
> cmd.OnAsm(func (*proxy.AsmCommand){ /* ... */ })
> cmd.OnLink(func (*proxy.LinkCommand){ /* ... */ })
> 
> // Execute the finalized command
> cmd.MustRun()
> ```
> 
> This way, you don't need the empty no-op implementations, and you can more directly target the changes you want to make.

I took another path that gets rid of this weird interface design that badly mimicked dynamic dispatch.
Now it's only about `proxy.MustParseCommand`, `proxy.ProcessComand` and `proxy.MustRunCommand`

internal/toolexec/proxy/parse.go
```diff
@@ -0,0 +1,53 @@
...
+	// We currently don't need to inject other tool calls, so we parse them as generic unsupported commands
+	default:
+		return &command{args: args}, nil
+	}
+}
+
+// MustParseCommand calls ParseCommand and exits on error
+func MustParseCommand(args []string) Command {
+	cmd, err := ParseCommand(args)
+	utils.ExitIfError(err)
```
RomainMuller, line 35: I reckon the "standard" practice in `Must*` functions is to panic in case of an error. The idea being that this would result from programmer/maintainer error, where the other case (IMHO corresponding to `utils.ExitIfError`) would be "incorrect use of the tool by the user".

internal/toolexec/proxy/parse.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return cmd
+}
+
+func parseCommandID(cmd string) (string, error) {
+	// It mustn't be empty
+	if cmd == "" {
```
RomainMuller, line 42: I feel this comment is obvious given what's just below :)

```suggestion
	if cmd == "" {
```

internal/toolexec/proxy/parse.go
```diff
@@ -0,0 +1,53 @@
...
+		return "", errors.New("unexpected empty command name")
+	}
+
+	// Take the base of the absolute path of the Go tool
+	cmd = filepath.Base(cmd)
+	// Remove the file extension if any
+	if ext := filepath.Ext(cmd); ext != "" {
```
RomainMuller, line 49: Is this to remove `.exe`, `.bat`, etc? Maybe mention this in the comment?

internal/toolexec/proxy/parse.go
```diff
@@ -0,0 +1,53 @@
...
+		cmd = strings.TrimSuffix(cmd, ext)
```
RomainMuller, line 50: Might be more efficient to do (modulo possible off-by-1 error):

```suggestion
		cmd = cmd[:len(ext)]
```

internal/toolexec/proxy/proxy.go
```diff
@@ -0,0 +1,120 @@
...
+		// paramPos is the index in args of the *value* provided for the parameter stored in the key
+		paramPos map[string]int
+		flags    commandFlagSet
+	}
+)
+
+const (
+	CommandTypeUnknown CommandType = "unknown"
+	CommandTypeCompile             = "compile"
+	CommandTypeLink                = "link"
```
RomainMuller, line 48: IIRC, you need to type these explicitly if you're not using `iota`:

```suggestion
	CommandTypeUnknown CommandType = "unknown"
	CommandTypeCompile CommandType = "compile"
	CommandTypeLink CommandType    = "link"
```

It might also actually be more efficient to use `int` based constants instead of strings, for example when performing `switch`.

internal/toolexec/proxy/proxy.go
```diff
@@ -0,0 +1,120 @@
...
+	"os/exec"
+	"path/filepath"
+)
+
+type (
+	// CommandType represents a Go compilation command type
+	// Currently support types are:
+	// - compile
+	// - link
+	// - unknown
```
RomainMuller, line 17: The "currently supported types" is more of an example IMHO... the actual list of supported stuff is best documented via the constants below.

```suggestion
	// CommandType represents a Go toolchain command type, such
	// as "compile", "link", etc...
```

internal/toolexec/proxy/proxy.go
```diff
@@ -0,0 +1,120 @@
...
+	CommandType string
+	// Command represents a Go compilation command
+	Command interface {
+		Args() []string
+		ReplaceParam(param string, val string) error
+		Stage() string
+		Type() CommandType
+	}
```
RomainMuller, line 25: Can the interface be named/extracted? (It does not necessarily need to be exported)

internal/toolexec/proxy/proxy.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (cmd *command) Stage() string {
+	return filepath.Base(filepath.Dir(cmd.flags.Output))
+}
+
+func (cmd *command) Type() CommandType {
+	return CommandTypeUnknown
+}
+
```
RomainMuller, line 117: Perhaps the `command` type should be renamed here if that's the "unknown" command only... And then maybe. `CommandTypeUnknown` should be `CommandTypeOther`? (Optional)

internal/toolexec/utils/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	"os/exec"
+	"strings"
+)
+
+// ExitIfError calls os.Exit(1) if err is not nil
+func ExitIfError(err error) {
+	if err != nil {
+		log.Printf("%v", err)
+		os.Exit(1)
+	}
```
RomainMuller, line 15: Style/Optional, although the `log.Printf` is unfortunate because it does not contain a trailing new line, which it should... Also `log.Fatalln` does what you're doing here... :P 

```suggestion
	if err == nil {
	  return
	}
	log.Fatalln(err)
```

internal/toolexec/utils/utils.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// GoBuild builds in provided dir and returns the work directory's true path
+func GoBuild(dir string, args ...string) (string, error) {
+	args = append([]string{"build", "-work", "-a", "-p", "1"}, args...)
```
RomainMuller, line 20: The documentation doesn't tell the whole story here...

- Implicitly says `-work`, but perhaps should mention you're responsible for cleaning up after it unless you wanna keep files around, still
- Doesn't mention it passes `-a`, and MAYBE this should be part of `args...` if the caller wants it
- Doesn't mention it passes `-p 1`, which I'd argue is not "the right way" to do it if there are concurrency issues

Hellzy: Right, the `-p` stuff should actually be removed - it's legacy from the bash POC.
I'll enrich the doc.

internal/toolexec/proxy/proxy.go
Hellzy: This was facilitating the transition from parsed string to command type representation.
I don't feel strongly about it though so I went ahead and used int instead, and added a small switch case to go from string to command type

internal/toolexec/tests/tests.sh
```diff
@@ -0,0 +1,19 @@
...
+cd $testDir
+go build -a -o proxy ./proxy
+
+for pkg in pkg_*
+do
+    # Inject $pkg into the base package
+    go build -a -o main -toolexec "$testDir/proxy/proxy $testDir/$pkg/cfg.yaml" ./base
+    out=$(./main)
+    # Make sure running the program yields the instrumented output
+    [ "$out" = "$pkg" ]
```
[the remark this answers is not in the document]
Hellzy: Alright, replaced it by `diff` with process substitution

internal/toolexec/tests/proxy/toolexec.go
```diff
@@ -0,0 +1,72 @@
...
+		dstAbs, _ := filepath.Abs(dst)
+		absReplace[srcAbs] = dstAbs
+	}
+	cfg.Replace = absReplace
+	return cfg, err
+}
+
+func main() {
+	log.SetFlags(0)
+	log.SetOutput(io.Discard)
```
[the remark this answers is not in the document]
Hellzy: Good catch, this wasn't supposed to stay here

internal/toolexec/proxy/flags.go
```diff
@@ -0,0 +1,89 @@
...
+	// Split the argument by its first `=` character if any, and check the
+	// syntax being used.
+	kv := strings.SplitN(arg, "=", 2)
+	option := kv[0]
+
+	flag, exists := flagSetValueMap[option]
+
+	if len(kv) == 2 {
+		// `-opt=val` syntax
+		value := kv[1]
```
RomainMuller, line 62: The `strings.Cut` function is a better fit for this use-case:

```suggestion
	option, value, hasValue := strings.Cut(arg, "=")
	flag, exists := flagSetValueMap[option]
	
	if hasValue {
		// `-opt=val` syntax
```

internal/toolexec/proxy/flags.go
```diff
@@ -0,0 +1,89 @@
...
+		shift = 1
+		if exists {
+			flag.SetString(value)
+		}
+	} else if nextArg == "" || len(nextArg) > 1 && nextArg[0] != '-' {
```
RomainMuller, line 67: How about:

```suggestion
	} else if !strings.HasPrefix(nextArg, "-") {
```

I find this is a bit more expressive, although it'll false-positive if the flag  was the very last argument of the list (and there is no "next", so the zero-value of string, which is "", would be used)

internal/toolexec/proxy/flags.go
```diff
@@ -0,0 +1,89 @@
...
+		value := nextArg
+		shift = 2
+		if exists {
+			switch flag.Kind() {
+			case reflect.String:
+				flag.SetString(value)
+			case reflect.Bool:
+				flag.SetBool(true)
+				shift = 1
+			}
```
RomainMuller, line 78:
```suggestion
			switch flag.Kind() {
			case reflect.String:
				flag.SetString(value)
			case reflect.Bool:
				flag.SetBool(true)
				shift = 1
			default:
				panic(fmt.Errorf("unsupported value kind: %s", flag.Kind()))
			}
```

internal/toolexec/proxy/flags.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	} else {
+		// `-opt` syntax (no value)
+		shift = 1
+		if exists && flag.Kind() == reflect.Bool {
+			flag.SetBool(true)
+		}
```
RomainMuller, line 85:
```suggestion
		if exists {
			if flag.Kind() != reflect.Bool {
				panic(fmt.Sprintf("missing value for %s flag", flag.Kind()))
			}
			flag.SetBool(true)
		}
```

RomainMuller (commented):
Alright this is largely good, I have a variety of style nitpicks but I like the concept.

On top of considering the style suggestions I'm making, I would like to have documentation comments on exported APIs that aren't entirely "obvious" (one example is you can't really tell what `Command.Stage()` returns without looking at the implementation, and even that doesn't tell you the full story unless you also know what gets into that code exactly).

internal/toolexec/processors/package.go
```diff
@@ -0,0 +1,227 @@
...
+	filepath.WalkDir(wDir, func(path string, d fs.DirEntry, err error) error {
+		if d.IsDir() || d.Name() != "importcfg" {
+			return nil
+		}
+		file, err := os.Open(path)
+		if err != nil {
+			return err
+		}
+		pkgReg.Combine(pkgRegiterFromImportCfg(file))
+		defer file.Close()
```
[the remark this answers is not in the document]
Hellzy: Nice catch

internal/toolexec/processors/package.go
```diff
@@ -0,0 +1,227 @@
...
+	return PackageInjector{
+		importPath: importPath,
+		pkgDir:     pkgDir,
+		buildFlags: flags,
+	}
+}
+
+// ProcessCompile visits a compile command, compiles the injected package
+// and includes the package dependency in the target package's importcfg
+func (i *PackageInjector) ProcessCompile(cmd *proxy.CompileCommand) {
```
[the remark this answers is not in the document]
Hellzy: This comment is actually specific to this implementation.
This implementation specifically compiles a package and injects it. Another Processor could do whatever, like replace a go file, add compile flags, add a go file, etc...

internal/toolexec/processors/package.go
```diff
@@ -0,0 +1,227 @@
...
+			log.Printf("err at entry: %v", err)
+			return err
+		}
+		if d.IsDir() || d.Name() != "importcfg" {
+			return nil
+		}
+
+		file, err := os.OpenFile(path, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0640)
+		if err != nil {
+			log.Printf("err browsing file dir: %v", err)
```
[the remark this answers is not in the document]
Hellzy: Right, this happens while browsing but the error should be specific to the operation that failed 👍

internal/toolexec/processors/state.go
```diff
@@ -0,0 +1,67 @@
...
+// It is used to keep track of whatever packages get built
+// This is saved to the disk in between toolexec calls in order
+// to keep some state from one call to another (mainly compile -> link)
+type State struct {
+	// mapping import : dependencies
+	Deps map[string]PkgRegister
+}
+
+func (s *State) serialize() ([]byte, error) {
+	var buf bytes.Buffer
```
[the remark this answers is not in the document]
Hellzy: `serialize()` ends up being a two liner that's only used in `SaveToFile`, so I just inlined it and removed the method

internal/toolexec/processors/package.go
```diff
@@ -0,0 +1,227 @@
...
+	err = filepath.WalkDir(filepath.Dir(cmd.Flags.Output), func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			log.Printf("err at entry: %v", err)
+			return err
+		}
+		if d.IsDir() || d.Name() != "importcfg" {
+			return nil
+		}
+
+		file, err := os.OpenFile(path, os.O_APPEND|os.O_RDWR|os.O_CREATE, 0640)
```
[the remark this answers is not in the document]
Hellzy: Good catch on the `O_CREATE`, shouldn't be there indeed.
For the rest... you had me at "JS doesn't like it" 😛.
Also yeah, Go seems to set `0o644`

internal/toolexec/processors/package.go
```diff
@@ -0,0 +1,227 @@
...
+func (i *PackageInjector) ProcessLink(cmd *proxy.LinkCommand) {
+	if cmd.Stage() != "b001" {
+		return
+	}
+	log.Printf("[%s] Injecting %s at link", cmd.Stage(), i.importPath)
+
+	// 1 - Read state from disk (created by ProcessCompile step)
+	log.Printf("====> Reading state from %s", ddStateFilePath)
+	state, err := StateFromFile(ddStateFilePath)
+	defer os.Remove(ddStateFilePath)
```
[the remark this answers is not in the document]
Hellzy: Maybe we should make it configurable? A mode where we remove ASM byproducts asap seems like something that we may want to allow. Maybe make it the default mode add an opt-in option to save the state?

internal/toolexec/examples/list-go-files.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"log"
+
+	"github.com/datadog/orchestrion/internal/toolexec/proxy"
+	"github.com/datadog/orchestrion/internal/toolexec/utils"
+)
+
+func ExampleListGofiles() {
+	args := []string{"/random/compile", "-trimpath", "randompath", "-p", "random", "-o", "/tmp/randomBuild/_pkg_.a", "-importcfg", "/tmp/random/b002/importcfg", "file1.go", "file2.go", "main.go"}
```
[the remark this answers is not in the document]
Hellzy: I tried the following (on top of changing the code a bit:
```
	// For the following toolexec invocation:
	// ./my-toolexec-program /random/compile -trimpath randompath -p random -o /tmp/randomBuild/_pkg_.a -importcfg /tmp/random/b002/importcfg file1.go file2.go main.go
	// We expect args to look something like this:
	// []string{"/random/compile", "-trimpath", "randompath", "-p", "random", "-o", "/tmp/randomBuild/_pkg_.a", "-importcfg", "/tmp/random/b002/importcfg", "file1.go", "file2.go", "main.go"}
```

I personally don't feel that it adds much more clarity, lmk what you think?

internal/toolexec/proxy/flags.go
Hellzy: Why not this then:
```suggestion
	} else if nextArg == "" || !strings.HasPrefix(nextArg, "-") {
```
Unless the `nextArg == ""` really bugs you? But I'd avoid having false positives

internal/toolexec/processors/package.go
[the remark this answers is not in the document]
RomainMuller: If these are _large_ objects, then yes, it might be a good thing to allow cleaning them up earlier rather than later...

My perspective on this is that the work tree will be cleaned up by the go toolchain unless `-work` is passed, and that would be the mechanism to allow the user to choose to keep intermediate files & directories

internal/toolexec/examples/list-go-files.go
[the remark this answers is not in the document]
RomainMuller: Yeah so the reason I was making this comment was to illustrate actual expected use of the API overall (so `os.Args`) while still letting folks know what kind of input is to be expected here... I guess the comment could be a lot lighter and merely include the contents of the `[]string`... 

But a the end of the day I don't think it changes things a lot... Maybe keep it the way you have it, but add a comment that says `// Normally os.Args()[1:]` or something? I dont know...

internal/toolexec/proxy/compile.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ReplaceFile is a convenience wrapper around Command.ReplaceParam to
+// replace a Go file by another in the cmd arguments
+func (cmd *CompileCommand) ReplaceFile(old, new string) error {
+	if !strings.HasSuffix(old, ".go") {
+		return fmt.Errorf("%s is not a Go file", old)
+	}
+	if !strings.HasSuffix(new, ".go") {
+		return fmt.Errorf("%s is not a Go file", new)
+	}
```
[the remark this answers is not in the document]
RomainMuller: Maybe we can do away with it now, and re-add it if it turns out to be painful without this helper?

internal/toolexec/proxy/flags.go
RomainMuller: This is redundant... if `nextArg == ""` then it does not have the `"-"` prefix.

internal/toolexec/processors/package.go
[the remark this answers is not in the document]
Hellzy: Fair enough. Removing the line 👍

internal/toolexec/examples/list-go-files.go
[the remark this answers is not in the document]
Hellzy: 81f14335aae745

internal/toolexec/proxy/flags.go
Hellzy: I think just checking for a non-empty next arg + no prefix works.
```
else if nextArg != "" && !strings.HasPrefix(nextArg, "-")

RomainMuller (approved):
Let's go!

PR #63, opened 2024-01-29
feat: configurable injection library (AOP-inspired)

RomainMuller:
### What does this PR do?

This PR introduces a new composable and declarative way to specify code injections in a way that is inspired from Aspect Oriented Programming (AOP), which are composed of:
1. Join-points, which specify where new behavior needs to be added
2. Advices, which specify what new behavior is to be added there

This includes a set of join-point operators and advice implementations which are designed to replicate Orchestrion's currently existing instrumentations, YAML configurations for each of these, a parser, and a Go code generator that allows removing the parsing stage from Orchestrion's start-up in cases when only the built-in configuration is used.

The new features are currently inert, as we will integrate them together with a `go -toolexec` proxy feature.

### Motivation

This streamlines authoring of code injections and allows coverage to scale faster than with hand-coded approach, while also allowing a low-code interface. This would allow customers to provide their own injection configurations for use with Orchestrion if they wanted to.

### Reviewer's Checklist

- [X] Changed code has unit tests for its functionality.


.github/workflows/generate.yml
RomainMuller, line 1: There's a few things produced by `go generate` in this repository, this workflow ensures PRs include up-to-date versions of these files.

_integration-tests/start.sh
```diff
@@ -0,0 +1,83 @@
...
+  for (( i=0; i<5; i++ )); do
+    curl -f "http://${DD_AGENT_HOST:-localhost}:8126/test/session/traces?test_session_token=${token}" \
+      -o "/output/traces.json"
+    if [[ "$(jq '. | length' < /output/traces.json)" == "0" ]]; then
+      echo "Agent returned no traces... retrying in 1 second..."
+      sleep 1
+    else
+      break
+    fi
+  done
```
RomainMuller, line 46: This loop is to avoid the tests being flakey... Turns out even without significant load, the fake agent may take a short moment to reach full consistency...

_integration-tests/tests/chi.v5/validation.json
```diff
@@ -4,7 +4,7 @@
 	{
 	    "name": "http.request",
 	    "service": "chi.router",
-	    "resource": "GET unknown",
+	    "resource": "GET /",
```
RomainMuller, line 7: This change is the result of upgrading `chi` to the latest patch release.

_integration-tests/tests/sql/validation.json
RomainMuller, line 1: Made indentation consistent in this file. No other changes.

_integration-tests/validator/validator.go
```diff
@@ -156,17 +157,28 @@ func main() {
...
+		if err != nil {
+			fmt.Printf("Failed to open traces file %q: %v", url.Path, err)
+			os.Exit(1)
+		}
+		defer file.Close()
+		if fakeTraces, err = assembleTraces(file); err != nil {
+			fmt.Printf("Failed to decode traces from %q: %v\n", url.Path, err)
+			os.Exit(1)
+		}
+	} else if resp, err := http.Get(*surl); err != nil {
```
RomainMuller, line 172: Added support for `file://` URLs there in case you want to just use a file that's on the local FS.

instrument/event/event.go
```diff
@@ -5,7 +5,7 @@
 
 package event
 
-//go:generate stringer -type=Event
+//go:generate go run golang.org/x/tools/cmd/stringer -type=Event
```
RomainMuller, line 8: This is to stop assuming `stringer` is installed "globally" and instead pick it up from `go.mod` via `tools.go`.

internal/injector/aspect/advice/code/dot_ast.go
```diff
@@ -0,0 +1,12 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+//go:generate go run ./generator -o dot_ast.proxies.go
```
RomainMuller, line 6: Generates proxies that essentially ensure we have the correct `.String() string` implementation to "move" existing nodes into the generated code. This allows the templates to traverse the AST transparently.

internal/injector/aspect/advice/code/generator/main.go
RomainMuller, line 1: This generator generates proxy types for all node types in `dave/dst`, generating the `dot_ast.proxies.go` file (via `go:generate` in `dot_ast.go`). The code there is pretty verbose but it's relatively straight foward.

internal/injector/aspect/advice/code/template.go
RomainMuller, line 1: This file centralizes the "go template --> DST nodes" operations. It tries to be as low overhead as possible when dealing with templates (in particular, it strives to parse them **exactly once**).

This one is likely interesting to review.

internal/injector/basiclit/basiclit.go
RomainMuller, line 1: This is a simple utility to "parse out" string values from AST nodes, as they're provided in their "go syntax" form (not really convenient). This could be replaced if there is a function somewhere in the standard library to "evaluate" such syntax, but I'm not aware of that?

internal/injector/builtin/generator/main.go
RomainMuller, line 1: This turns `.yml` configurations into go code so we don't have to parse YAML at the start of `orchestrion` all the time, as this could end up being quite pricey. It also saves us from having to embed the YAML into the binary, which would unnecessarily increase the file size.

internal/injector/builtin/integration_test.go
```diff
@@ -0,0 +1,95 @@
...
+				PreserveLineInfo: true,
+			})
+			require.NoError(t, err)
+
+			files, err := filepath.Glob(path.Join(pkgDir, "*.go"))
+			require.NoError(t, err)
+			for _, filename := range files {
+				t.Run(path.Base(filename), func(t *testing.T) {
+					res, err := inj.InjectFile(filename, map[string]string{"httpmode": "wrap"})
+					require.NoError(t, err)
```
RomainMuller, line 56: This test is a little setup-heavy (might refactor some of that later), but here is where "the business" happens.

internal/injector/builtin/testdata/go.mod
RomainMuller, line 1: This file isolates the `testdata` files from the surrounding module. These are reference files from the built-in configuration integration test... They're produced from running the `Injector` against the files in `/samples`.

internal/injector/code/code.go
RomainMuller, line 1: This is a trivial interface to provide support for YAML to go code transformation. I'd have hidden it, but Go visibility semantics don't offer me the flexibility I'd need there, and I wanted to have "guaranteed" implementation, which I wouldn't have had with a last-minute type assertion (and anyhow this would still require the `AsCode` function to be exported, unfortunately).

internal/injector/injector.go
RomainMuller, line 1: This file is where the magic starts. There's callback-based AST traversal in there (via `dstutil` which is worthy of review.

The `decorator.Load` stuff is one where there might be performance gains to be had from streamlining or smartly applying cache (especially if the intent is to instrument a package tree), but I've intentionally left that "for later" once we have concrete use-cases.

internal/injector/injector_test.go
```diff
@@ -0,0 +1,113 @@
...
+				child.Stdout = os.Stdout
+				child.Stderr = os.Stderr
+				return child.Run()
+			}
+			require.NoError(t, run("go", "mod", "tidy"), "failed to run go mod tidy")
+			require.NoError(t, run("go", "mod", "download"), "failed to run go mod download")
+
+			injector, err := injector.New(path.Dir(filename), tc.Options)
+			require.NoError(t, err)
+			res, err := injector.InjectFile(filename, nil)
```
RomainMuller, line 91: "The business" is here

internal/injector/node/chain.go
RomainMuller, line 1: This is an interesting little contraption to make sure all the surrounding context of a given DST node is available to code templates, etc... No rocket science, but still interesting to look at.

internal/injector/singleton/singleton.go
RomainMuller, line 1: The singleton is used to facilitate polymorphic deserialization from YAML, and assumes objects are encoded as:

```yml
<discriminator>: <properties>
```

Hence, this ensures it is provided a singleton mapping (a mapping with exactly 1 key-value pair in it) and extracts the key (discriminator) & value (properties) from there.

internal/injector/typed/context.go
RomainMuller, line 1: I wish this was in the standard library 🫣

tools.go
RomainMuller, line 1: Ensures `go:generate` tools are always available so you can `go generate ./...` on a fresh checkout without having to know what needs installing first.

Hellzy (approved):
Talk about reading material 😅 

As I mentioned, given the sheer size of this PR I focused on understanding what's happening rather than trying to redesign/challenge it. The overall architecture makes sense to me so I feel confident approving it 👍

PR #66, opened 2024-03-11
feat: ensure version run matches go.mod requirements

RomainMuller:
### What does this PR do?

This PR introduces a feature that'd allow replacing the current process with a new one that runs from `go run github.com/datadog/orchestrion` so that it uses the same version as is declared in `go.mod`, allowing `orchestrion` to be installed using `go install` while still enforcing reproductible builds via `go.mod` locking.

### Reviewer's Checklist

### DIFF
--- a/internal/ensure/requiredversion.go
+++ b/internal/ensure/requiredversion.go
@@ -10,6 +10,8 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
+	"runtime"
 	"syscall"
 
 	"github.com/datadog/orchestrion/internal/version"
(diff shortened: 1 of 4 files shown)


internal/ensure/requiredversion.go
```diff
@@ -0,0 +1,124 @@
...
+	goBin, err := exec.LookPath("go")
+	if err != nil {
+		return fmt.Errorf("failed to resolve go from PATH: %w", err)
+	}
+
+	args := make([]string, len(osArgs)+2)
+	args[0] = goBin
+	args[1] = "run"
+	args[2] = orchestrionPkgPath
+	copy(args[3:], osArgs[1:])
```
Hellzy, line 92: This assumes that `requiredVersion` gets called with len(osArgs) >= 1, otherwise we're writing out of bounds. Since this seems to be a strong assertion given that we expect to never receive an empty slice (probably copied from `os.Args`), why not assert it in the code with a panic?

Hellzy (commented):
WDYT of running those tests in the CI as well?

RomainMuller:
> WDYT of running those tests in the CI as well?

What do you mean? They _are_ run as part of `go test`?

PR #70, opened 2024-03-27
feat: decorate output of `toolexec [...] -V=full`

RomainMuller:
The go build toolchain calls tooling with `-V=full` to compute the full build ID, allowing for invalidation of cached build artifacts when tooling changes.

The contract for this option is to print a single line of content, which will be embedded in the build ID. We are hence intercepting these in the `-toolexec` flow and appending the Orchestrion version and ruleset information to the output.

---

Additionally, this removes ability to force build (pass `-a`) and override the `GOCACHE`, since these measure ought to make these two features redundant.

### DIFF
--- a/internal/goproxy/proxy.go
+++ b/internal/goproxy/proxy.go
@@ -15,40 +15,29 @@ import (
 )
 
 type config struct {
-	differentCache bool
-	forceBuild     bool
-	toolexecArgs   string
+	toolexec string
 }
 
 type Option func(*config)
 
-const goCacheVar = "GOCACHE"
-
-// WithForceBuild forces a call to Run() to build all go files and dependencies
-// when wrapping a `go build` command. It effectively adds the `-a` flag to the build
-func WithForceBuild() Option {
-	return func(c *config) {
-		c.forceBuild = true
-	}
-}
-
-// WithDifferentCache forces a call to Run() to override the GOCACHE directory
-// A temporary directory is instead created and all build artifacts are cached there
-func WithDifferentCache() Option {
-	return func(c *config) {
-		c.differentCache = true
-	}
-}
-
 // WithToolexec forces a call to Run() to build with the -toolexec option when
 // wrapping a build command
-func WithToolexec(args []string) Option {
-	var sb strings.Builder
-	return func(c *config) {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -32,14 +35,25 @@ func main() {
 		fmt.Println(version.Tag)
 		return
 	case "go":
-		err := goproxy.Run(args, goproxy.WithForceBuild(), goproxy.WithDifferentCache())
+		orchestrion, err := os.Executable()
 		if err != nil {
+			orchestrion = os.Args[0]
```
Hellzy, line 40: How about using `filepath.Abs` in that case? If orchestrion is called from a relative path it may cause issues combined with toolexec down the road

internal/goproxy/proxy.go
```diff
@@ -15,76 +15,62 @@ import (
...
-}
-
 // WithToolexec forces a call to Run() to build with the -toolexec option when
 // wrapping a build command
-func WithToolexec(args []string) Option {
-	var sb strings.Builder
-	return func(c *config) {
-		for _, arg := range args {
-			sb.WriteString(fmt.Sprintf("%s ", arg))
+func WithToolexec(args ...string) Option {
```
Hellzy, line 25: How about the following since we at least expect the toolexec binary to be provided
```suggestion
func WithToolexec(tool string, args ...string) Option {
```

Hellzy (commented):
Seems good to me, left a couple suggestions

main.go
RomainMuller: `os.Executable()` kind of does that already, but in a different way (and TBH it makes me a little uneasy when I read that code... but I suppose it was written by people smarter than me).

It nevertheless is a good idea. That should be an _extremely_ cold path in normal circumstances anyway!

Hellzy (approved):

RomainMuller:
The go build toolchain calls tooling with `-V=full` to compute the full build ID, allowing for invalidation of cached build artifacts when tooling changes.

The contract for this option is to print a single line of content, which will be embedded in the build ID. We are hence intercepting these in the `-toolexec` flow and appending the Orchestrion version and ruleset information to the output.

PR #75, opened 2024-04-08
feat: -toolexec with aspect weaving

RomainMuller:
Adds aspect weaving to the `-toolexec` stack:
- Transforms code as necessary ahead of running compilers
- Injects new library dependencies as necessary
  - a `link.deps` text file is added to the `_pkg_.a` files that contains any/all injected import paths, so the linker is able to provide them

This is `GOCACHE` friendly and relies on `GOCACHE` to de-duplicate build artifacts (on an empty cache, some objects may be built more than once, but this should be a marginal cost).


internal/toolexec/processors/aspect/linkdeps/linkdeps.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+
+		if l.deps == nil {
+			l.deps = make(map[string]struct{}, 1)
+		}
+		l.deps[line] = struct{}{}
```
Hellzy, line 81:
```suggestion
		l.Add(line)
```

internal/toolexec/processors/aspect/linkdeps/linkdeps.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Add registers a new import path in this LinkDeps instance.
+func (l *LinkDeps) Add(importPath string) {
+	if l.deps == nil {
+		l.deps = make(map[string]struct{})
+	}
+	l.deps[importPath] = struct{}{}
+}
+
+// Direct returns all import paths registered in this LinkDeps instance that are
```
Hellzy, line 93: Seems to me it returns them all. Do you mean that LinkDeps only hold direct deps?

internal/toolexec/processors/aspect/linkdeps/linkdeps.go
```diff
@@ -0,0 +1,146 @@
...
+	switch hdr := strings.TrimSpace(line); hdr {
+	case headerV1:
+		return parseV1(rd)
+	default:
+		err = fmt.Errorf("unsupported data format %q, a newer Orchestion release may be required", hdr)
+		return
+	}
+}
+
+// parseV1 parses the contents of V1 link.deps files.
```
Hellzy, line 58: What's the versioning for?

internal/toolexec/processors/aspect/oncompile.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if linkDeps.Empty() {
+		// There are no synthetic dependencies, so we don't need to write an updated importcfg or add
+		// extra objects in the output file.
+		return nil
+	}
+
+	if regUpdated {
+		// Creating updated version of the importcfg file, with new dependencies
+		if err := writeUpdatedImportConfig(reg, cmd.Flags.ImportCfg); err != nil {
```
Hellzy, line 114: Should we consider writing a seperate synthetic importcfg and use it with `-importcfg` at some point?

internal/toolexec/processors/aspect/weaver.go
```diff
@@ -0,0 +1,10 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package aspect
+
+type Weaver struct {
```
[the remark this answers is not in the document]
RomainMuller: Yeah you have to think of the package as "the unit", and then it all works neatly...

internal/toolexec/processors/aspect/linkdeps/linkdeps.go
RomainMuller: This is somewhat future-proofing; so that if we want to change the content/format of this, we have an easy & early detection mechanism which affords us outputting an appropriate error message...

internal/toolexec/processors/aspect/oncompile.go
RomainMuller: This is actually what's happening, except in order to simplify things a bit, this renames the original to `importcfg.original` then writes the updated version in-place.

internal/toolexec/processors/aspect/linkdeps/linkdeps.go
RomainMuller: Yeah it holds only direct dependencies at this point... I had a previous iteration that had both direct & transitive, but it turned out to be a whole lot of duplicated effort that couldn't be avoided... so I simplified...

internal/toolexec/processors/aspect/onlink.go
```diff
@@ -0,0 +1,99 @@
...
+	defer file.Close()
+
+	if _, err := reg.WriteTo(file); err != nil {
+		return fmt.Errorf("writing updated %q: %w", cmd.Flags.ImportCfg, err)
+	}
+
+	return nil
+}
+
+func archiveData(archive, entry string) (io.Reader, error) {
```
Hellzy, line 75: Nit: this func name kinda got me confused while reading the code since my first thought was that it actually would archive some data. `dataFromArchive()` or `readArchiveData()` may be a bit more explicit.
Or a short comment.

Hellzy (approved):
I really like that you left comments to  explain the "non-trivial" branching cases, it made the review so much easier. Thanks 🙏

internal/toolexec/processors/aspect/onlink.go
RomainMuller: Good point... I think `read` prefix would help indeed!

PR #79, opened 2024-04-10
feat: Go command flags propagation

Hellzy:
### What does this PR do?

Introduce the `goflags` package that backtracks through the process stack to find the parent go command invocation
and parse its flags. These flags can in turn be propagated to the go commands that are run by the toolexec proxy.

Two cases can happen:
- `orchestrion go <cmd> <args>` invocation: we pass through the go proxy to re-invoke the command in the form `go build <args> -toolexec <...>`=> fallback to the case bellow
- `go <cmd> <args> -toolexec <...>`: we backtrack through the processes to find the whole go command and retrieve its flags. They can then be added to the commands we decide to run in the weaver

Go command flags can be retrieved by calling `goproxy.Flags()`, which takes care of the backtracking and the parsing.

### Reviewer's Checklist


internal/goflags/flags.go
```diff
@@ -0,0 +1,140 @@
...
+	return flags
+}
+
+// Flags return the top level go command flags
+func Flags() (CommandFlags, error) {
+	var err error
+	if flags == nil {
+		*flags, err = parentGoCommandFlags()
+		flags.Trim("-toolexec", "-o")
+	}
```
RomainMuller, line 90: This could instead be achieved with a `sync.Once` (but I don't care that much)

internal/goflags/flags.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return *flags, err
+}
+
+func isOption(str string) bool {
+	return len(str) > 0 && str[0] == '-'
+}
+
+func isAssigned(str string) bool {
+	return regexp.MustCompile(".+=.+").MatchString(str)
```
RomainMuller, line 100: You should store the compiled regex in a package-scoped variable, so that you don't pay the price of parsing it over and over again.

Next, I think this regular expression is overkill, as `strings.Cut` would effectively give you the same signal, for probably cheaper?

I also wonder if there is a scenario where you can have whitespace in there instead of the `=` (especially in case where the process was started by `exec[v[e|p]]`?

internal/goflags/flags_test.go
```diff
@@ -0,0 +1,87 @@
...
+			for _, flag := range tc.remove {
+				require.NotContains(t, tc.flags.Long, flag)
+				require.NotContains(t, tc.flags.Short, flag)
+			}
+		})
+
+	}
+}
+
+func TestParse(t *testing.T) {
```
RomainMuller, line 41: This should also cater for the fancy arguments that pass down to the linker, etc...

- `-gcflags '-N -L'` 
- `-ldflags '-extldflags "-lm -lstdc++ -static"'`

I suspect these wouldn't work properly with the current logic...

internal/goflags/flags.go
Hellzy: Yes usually assigned flags can be whitespaced as well but in this case they'll fall back to being parsed as regular long flags

internal/goflags/flags.go
```diff
@@ -0,0 +1,169 @@
...
+	specialFlags = map[string]struct{}{
+		"-asmflags": {}, "-gccgoflags": {}, "-gcflags": {}, "-ldflags": {},
+	}
+	shortFlags = map[string]struct{}{
+		"-asan": {}, "-cover": {}, "-linkshared": {}, "-modcacherw": {}, "-msan": {}, "-race": {}, "-trimpath": {},
+	}
+	longFlags = map[string]struct{}{
+		"-covermode": {}, "-coverpkg": {}, "-buildmode": {}, "-buildvcs": {}, "-compiler": {}, "-mod": {},
+		"-modfile": {}, "-overlay": {}, "-pgo": {}, "-pkgdir": {}, "-tags": {},
+	}
```
RomainMuller, line 35: I wouldn't be afraid of using vertical space in there... Especially as this would make future updates more readable (also these should probably be sorted alphanumerically):

```suggestion
	specialFlags = map[string]struct{}{
		"-asmflags":   {}, // Flags passed through to the assembly
		"-gccgoflags": {}, // Flags passed through to the gccgo compiler
		"-gcflags":    {}, // Flags passed through to the gc compiler
		"-ldflags":    {}, // Flags passed through to the linker
	}
	shortFlags = map[string]struct{}{
		"-asan":       {}, // Enables address sanitizer
		"-cover":      {}, // Enables coverage colletion
		"-linkshared": {}, // Build code that links against shared libraries
		"-modcacherw": {}, // Keep module cache files read-write
		"-msan":       {}, // Enable memory sanitizer
		"-race":       {}, // Enable data race detection
		"-trimpath":   {}, // Remove all file system paths from the resulting executable
	}
	longFlags = map[string]struct{}{
		"-buildmode": {}, // Set build mode
		"-buildvcs":  {}, // Whether to stamp binaries with version control information
		"-compiler":  {}, // Select what compiler to use
		"-covermode": {}, // Set coverage mode
		"-coverpkg":  {}, // Set list of packages to collect coverage for
		"-mod":       {}, // Set module download mode
		"-modfile":   {}, // Set module file
		"-overlay":   {}, // Set overlay file
		"-pgo":       {}, // Set profile-guided optimization profile file
		"-pkgdir":    {}, // Set package install & load directory
		"-tags":      {}, // Set build tags
	}
```

internal/goflags/flags.go
```diff
@@ -0,0 +1,168 @@
...
+	"sync"
+
+	"github.com/datadog/orchestrion/internal/goproxy"
+	"github.com/shirou/gopsutil/v3/process"
+)
+
+// CommandFlags represents the flags provided to a go command invocation
+type CommandFlags struct {
+	Long    map[string]string
+	Short   []string
```
RomainMuller, line 20: I wonder if these should be a set (`map[string]struct{}`)

internal/goflags/flags.go
```diff
@@ -0,0 +1,168 @@
...
+// Trim removes the specified flags and their value from the long and short flags
+func (f CommandFlags) Trim(flags ...string) {
+	for _, flag := range flags {
+		delete(f.Long, flag)
+		for i, fl := range f.Short {
+			if fl == flag {
+				f.Short = append(f.Short[:i], f.Short[i+1:]...)
+				break
+			}
+		}
```
RomainMuller, line 47: See comment above about using a set... Here... the flag could be present multiple times in `f.Short` and you'd only trim the first one...

internal/goflags/flags.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// Slice returns the command flags as a string slice
+// - long flags are returned as a string of the form '-flagName="flagVal"'
+// - short flags are returned as a string of the form '-flagName'
+// - unknow flags and values are ignored
+func (f CommandFlags) Slice() []string {
+	flags := make([]string, 0, len(f.Long)+len(f.Short))
+	for flag, val := range f.Long {
+		flags = append(flags, fmt.Sprintf("%s=\"%s\"", flag, val))
```
RomainMuller, line 58: I'd use `%q` to make sure nested quotes are properly handled...

```suggestion
		flags = append(flags, fmt.Sprintf("%s=%q", flag, val))
```

internal/goflags/flags.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	for _, flag := range f.Short {
+		flags = append(flags, flag)
+	}
```
RomainMuller, line 62: Assuming you don't use a set... this can be written more efficiently as:
```suggestion
	flags = append(flags, f.Short...)
```

internal/goflags/flags.go
```diff
@@ -0,0 +1,168 @@
...
+	once.Do(func() {
+		flags, flagsErr = parentGoCommandFlags()
+	})
+	return flags, flagsErr
+}
+
+func isAssigned(str string) bool {
+	_, _, ok := strings.Cut(str, "=")
+	return ok
+}
```
RomainMuller, line 110: Perhaps this should also check that `strings.HasPrefix(str, "-")`

internal/goflags/flags.go
Hellzy: Good point

internal/goflags/flags_test.go
Hellzy: Added test cases for special flags

internal/goflags/flags.go
```diff
@@ -0,0 +1,182 @@
...
+// CommandFlags represents the flags provided to a go command invocation
+type CommandFlags struct {
+	Long    map[string]string
+	Short   map[string]struct{}
+	Unknown []string // flags we don't process but store anyway
+}
+
+var (
+	// specialFlags are long Go command flags that accept other flags as parameters
+	specialFlags = map[string]struct{}{
```
RomainMuller, line 26: What's the impact of this? Should these be merged with `longFlags`? It feels like they're handled identically... if the difference is only in what kind of values they get... this is not relevant here?

internal/toolexec/processors/aspect/resolve.go
```diff
@@ -72,7 +74,15 @@ func resolvePackageFiles(importPath string) (map[string]string, error) {
...
-		if err := exec.Command("go", "build", "-toolexec", toolexec, "--", importPath).Run(); err != nil {
+		// Retrieve original go command flags and pass them along
+		flags, err := goflags.Flags()
+		if err != nil {
+			return nil, fmt.Errorf("retrieving go command flags: %v", err)
+		}
+		flagsSlice := flags.Slice()
+		args := append([]string{"go", "build"}, flagsSlice...)
+		args = append(args, "-toolexec", toolexec, "--", importPath)
+		if err := exec.Command("go", args...).Run(); err != nil {
```
RomainMuller, line 85: This is turning into `go go build ...`, isn't it?

PR #80, opened 2024-04-10
chore: improve integration tests

RomainMuller:
Allows re-use of the Go object and module cache, which significantly improves the overall integration test suite duration.


Hellzy (approved):
Nice speed-up 👍

PR #83, opened 2024-04-11
feat: logging framework

RomainMuller:
Introduced a minimal logging framework to ease up troubleshooting operations.
    
This is configured via environment variables to avoid interfering with
command line argument handling, in particular with the `toolexec`
subcommand, which is pretty sensitive:
- `ORCHESTRION_LOG_LEVEL=[NONE|ERROR|WARN|INFO|DEBUG|TRACE]` sets the
  log level to use; it defaults to `NONE` unless `ORCHESTRION_LOG_FILE`
  is set, in which case it defaults to `WARN`.
- `ORCHESTRION_LOG_FILE=path/to/file.log` sets the log file to use; the
  file is appended to, if it already exists. When not specified, log
  output goes to `STDERR`. If the file name contains `"$PID"`, that will
  be replaced with the current process ID.
    
The logger uses `syscall.Flock` to avoid causing multiple processes
writing to the same file causing data corruption (generally leading to
an unreadable mess).


main.go
```diff
@@ -99,25 +104,30 @@ func main() {
 					}
 				}
 			}
+			log.Tracef("Appending orchestrion information to otuput: orchestrion@%s,%s\n", versionString, builtin.Checksum)
```
Hellzy, line 107:
```suggestion
			log.Tracef("Appending orchestrion information to output: orchestrion@%s,%s\n", versionString, builtin.Checksum)
```

internal/toolexec/aspect/oncompile.go
```diff
@@ -135,10 +144,12 @@ func (w Weaver) OnCompile(cmd *proxy.CompileCommand) error {
 func writeUpdatedImportConfig(reg importcfg.ImportConfig, filename string) (err error) {
 	const dotOriginal = ".original"
 
+	log.Tracef("Backing up original %q\n", filename)
```
Hellzy, line 147: Any reason this log is `Tracef` when all others are `Debugf`?

internal/log/log.go
```diff
@@ -0,0 +1,117 @@
...
+	if value != "" {
+		if _, found := context[key]; !found {
+			contextKeys = append(contextKeys, key)
+		}
+		context[key] = value
+	} else {
+		delete(context, key)
+		for i := 0; i < len(contextKeys); {
+			if contextKeys[i] == key {
+				contextKeys = append(contextKeys[:i], contextKeys[i+1:]...)
```
Hellzy, line 65: Not a big deal but you can actually break out early and save some iterations here, since the other branch guarantees that a given context key is only inserted once.

Hellzy (approved):

PR #90, opened 2024-04-23
fix: `append-args` breaks on ellipsis

### LINKED ISSUE #88: [BUG] Ellipses on varargs methods fails compilation
Version of orchestrion: v0.7.0-dev.1

For example a method with the prototype as follows:

```go
func test(x string, y ...string)
```

and a call done like this:

```go
y := []string{"y1", "y2"}
test("x", y...)
```

will be instrumented like this:

```go
y := []string{"y1", "y2"}
test("x", y, interceptor...)
```

Which fails the compilation. Something like this should work:

```go
y := []string{"y1", "y2"}
test("x", append(interceptor, y...)...)
```

RomainMuller:
When an ellipsis is used on a call that is targeted by an `append-args` advice, the new arguments were simply added to the list of values, which resulted in invalid code.

To address this, the `append-args` advice now injects an IIFE that correctly generates a new slice with the additional arguments injected in. The use of an IIFE allows removing problems if the variadic argument is provided as a function call (calling that multiple times could result in duplicated side-effects, etc...).

Fixes #88

### DIFF
--- a/internal/injector/aspect/advice/call.go
+++ b/internal/injector/aspect/advice/call.go
@@ -10,7 +10,9 @@ import (
 	"fmt"
 
 	"github.com/datadog/orchestrion/internal/injector/aspect/advice/code"
+	"github.com/datadog/orchestrion/internal/injector/aspect/join"
 	"github.com/datadog/orchestrion/internal/injector/node"
+	"github.com/datadog/orchestrion/internal/injector/typed"
 	"github.com/dave/dst"
 	"github.com/dave/dst/dstutil"
 	"github.com/dave/jennifer/jen"
@@ -18,11 +20,14 @@ import (
 )
 
 type appendArgs struct {
+	typeName  join.TypeName
 	templates []code.Template
 }
 
-func AppendArgs(templates ...code.Template) *appendArgs {
-	return &appendArgs{templates}
+// AppendArgs appends arguments of a given type to the end of a function call. All arguments must be
+// of the same type, as they may be appended at the tail end of a variadic call.
+func AppendArgs(typeName join.TypeName, templates ...code.Template) *appendArgs {
+	return &appendArgs{typeName, templates}
 }
 
 func (a *appendArgs) Apply(ctx context.Context, chain *node.Chain, csor *dstutil.Cursor) (bool, error) {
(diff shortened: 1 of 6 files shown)


internal/injector/aspect/advice/call.go
```diff
@@ -40,13 +45,65 @@ func (a *appendArgs) Apply(ctx context.Context, chain *node.Chain, csor *dstutil
...
-	call.Args = append(call.Args, newArgs...)
+	if !call.Ellipsis {
+		call.Args = append(call.Args, newArgs...)
+		return true, nil
+	}
+
+	// The function call has an ellipsis, so we need to append our new arguments to the last argument,
+	// which is a slice. To do so, we need to provision a new slice of the right type and size, append
+	// all the relevant data in there, and then replace the last argument with the new slice.
+	lastIdx := len(call.Args) - 1
```
Julio-Guerra, line 56: what allows you to know that `len(call.Args) - 1 >= 0` so the following code never panics?

internal/injector/aspect/advice/call.go
```diff
@@ -40,13 +45,65 @@ func (a *appendArgs) Apply(ctx context.Context, chain *node.Chain, csor *dstutil
...
+					&dst.ReturnStmt{Results: []dst.Expr{
+						&dst.CallExpr{
+							Fun: dst.NewIdent("append"),
+							Args: append(
+								append(
+									make([]dst.Expr, 0, len(newArgs)+1),
+									dst.NewIdent("opts"),
+								),
+								newArgs...,
+							),
```
Julio-Guerra, line 83: isn't this equivalent?
```suggestion
							Args: append(
								make([]dst.Expr, 0, len(newArgs)+1),
								dst.NewIdent("opts"),
								newArgs...,
							),
```

append being variadic, you can merge the two together it seems to me

internal/injector/aspect/advice/call.go
```diff
@@ -40,13 +45,65 @@ func (a *appendArgs) Apply(ctx context.Context, chain *node.Chain, csor *dstutil
...
+	if !call.Ellipsis {
+		call.Args = append(call.Args, newArgs...)
+		return true, nil
+	}
+
+	// The function call has an ellipsis, so we need to append our new arguments to the last argument,
+	// which is a slice. To do so, we need to provision a new slice of the right type and size, append
+	// all the relevant data in there, and then replace the last argument with the new slice.
+	lastIdx := len(call.Args) - 1
+	call.Args[lastIdx] = &dst.CallExpr{
```
Julio-Guerra, line 57: So IIUC, this block is trying to replace something like `opts...` by `append(opts, newArgs)...`, in which case I suggest we enforce copying `opts` instead with the same trick you used in the code below by replacing by `append(make([]<type of opts>, 0, len(opts)+1), opts, newArgs)...`

internal/injector/aspect/advice/call.go
```diff
@@ -40,13 +45,65 @@ func (a *appendArgs) Apply(ctx context.Context, chain *node.Chain, csor *dstutil
...
+			},
+			Body: &dst.BlockStmt{
+				List: []dst.Stmt{
+					&dst.ReturnStmt{Results: []dst.Expr{
+						&dst.CallExpr{
+							Fun: dst.NewIdent("append"),
+							Args: append(
+								append(
+									make([]dst.Expr, 0, len(newArgs)+1),
+									dst.NewIdent("opts"),
```
Julio-Guerra, line 80: how do you know the identifier used in the original func call is opts? 🤔 Shouldn't you reuse the argument's AST instead in order to retrieve the identifier?

internal/injector/testdata/injector.yml
Julio-Guerra, line 1: It seems the Go test going with it is missing?

Julio-Guerra (changes requested):

internal/injector/aspect/advice/call.go
RomainMuller: The call has an ellipsis, so it has at least 1 argument... If not, the code includes a syntax error and won't compile anyway...

internal/injector/aspect/advice/call.go
RomainMuller: As far as I know, Go syntax only allows for a variadic argument to be provided EITHER:
- as a list of values
- as a splat/ellipsis expression

You can't mix both?

internal/injector/aspect/advice/call.go
RomainMuller: This is wrapped in an IIFE which receives the original options variadically, implying a slice copy already.

internal/injector/aspect/advice/call.go
RomainMuller: This is an argument provided to the IIFE, and I chose the name for it...

internal/injector/testdata/injector.yml
RomainMuller: See in `injector.yml`

internal/injector/aspect/advice/call.go
Julio-Guerra: Shame on me - closing this comment xD

eliottness:
I can confirm that this fix #88

internal/injector/testdata/injector.yml
RomainMuller: This is a scenario that is run by the test harness, which has not changed.

Julio-Guerra: Was it passing despite this PR fixing the compilation issue?

Julio-Guerra (approved):

PR #93, opened 2024-04-24
feat: required version check is now non-failing

RomainMuller:
In order to simplify the onboarding experience, it is no longer a hard requirement to have orchestrion listed in `go.mod`.

If present in `go.mod`, that version will be respected. If not, a distinctive warning will be printed in the `go version` output, and an `orchestrion pin` command allows for easy, automated insertion of the `orchestrion` dependency (adding a `orchestrion.tool.go` file and running the adequate `go get` command for you).


main.go
```diff
@@ -44,29 +46,65 @@ func main() {
...
+			tools.HeaderComment("//go:build tools")
+			tools.Anon(orchestrionModule)
+
+			file, err := os.OpenFile("orchestrion.tool.go", os.O_CREATE|os.O_WRONLY|os.O_EXCL, 0o644)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Unable to create tools.go file: %v\n", err)
+				os.Exit(1)
 			}
+			defer file.Close()
+			if err := tools.Render(file); err != nil {
```
Hellzy, line 82: We also may want to remove the file in case of error here

Hellzy (approved):

PR #96, opened 2024-04-25
feat: automatic pinning of orchestrion

### LINKED ISSUE #87: [BUG] `orchestrion version` does not respawn to `go.mod` version
**Version of orchestrion**
0.7.0-dev.1

**Describe what happened:**
Running `orchestrion version` shows the version installed, not the version from `go.mod` as expected.

### LINKED ISSUE #95: [BUG] error when resolving woven dependency
I was trying the new quick start mode with go install and it ends up in the following error:

```
~/d/appsec-go-test-app (main|✔) $ ./orchestrion go build -v .
github.com/gorilla/mux
# github.com/gorilla/mux
2024/04/24 22:34:07 resolving woven dependency on github.com/datadog/orchestrion/instrument: running ["go" "list" "-toolexec" "\"/Users/julio.guerra/dd/appsec-go-test-app/orchestrion\" \"toolexec\"" "-json" "-deps" "-export" "--" "github.com/datadog/orchestrion/instrument"]: exit status 1`
```

You can reproduce it on the appsec-go-test-app with:

```shell
$ GOBIN=$PWD go install github.com/datadog/orchestrion@main
$ ./orchestrion go build -v . 
```

RomainMuller:
This adds a new feature which will automatically create the `orchestrion.tool.go` file, then run `go get` to explicitly install the current version of `orchestrion` in the `go.mod` file, then run `go mod tidy` to ensure the `go.mod` file is clean.

This is done only if the required version check fails (indicating the `go.mod` file did not contain a version for `orchestrion`). A new environment variable is set so that child processes skip the check if it was successful for their parent, saving time & risk for errors.

Upon automatic pinning, the following message is print out to STDERR (formatting varies depending on whether it's a PTY or not, etc... and if it's a PTY, the box width is ajdusted to fit the viewport):
```
╭──────────────────────────────────────────────────────────────────────────────╮
│                                                                              │
│  Warning: github.com/datadog/orchestrion is not present in your go.mod       │
│  file.                                                                       │
│  In order to ensure build reliability and reproductibility, orchestrion      │
│  will now add itself in your go.mod file by:                                 │
│                                                                              │
│      1. creating a new file named orchestrion.tool.go                        │
│      2. running go get github.com/datadog/orchestrion@v0.7.0-dev.2           │
│      3. running go mod tidy                                                  │
│                                                                              │
│  You should commit the resulting changes into your source control system.    │
│                                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

Fixes #95
Fixes #87 

---

In addition to this, a new `orchestrion warmup` command will pre-build the instrumented standard library as well as all packages `orchestrion` may be injecting into compiled code, speeding up future compilations.


internal/toolexec/aspect/resolve.go
```diff
@@ -19,79 +19,61 @@ import (
...
+	}
+	var items []listItem
+	dec := json.NewDecoder(&stdout)
+	for {
+		var item listItem
+		if err := dec.Decode(&item); err == io.EOF {
+			break
+		} else if err != nil {
+			return nil, fmt.Errorf("parsing `go list` output: %w", err)
 		}
```
eliottness, line 56: No `else` plz.
```suggestion
		}
		if err != nil {
			return nil, fmt.Errorf("parsing `go list` output: %w", err)
		}
```

internal/toolexec/aspect/resolve.go
eliottness, line 1: I am trying and failing to see how is it related to orchestrion pinning

internal/toolexec/aspect/resolve.go
```diff
@@ -19,79 +19,61 @@ import (
...
-		// Retrieve original go command flags and pass them along
-		args, err = prepareGoCommandArgs("build", "-toolexec", toolexec, "--", importPath)
-		if err != nil {
-			return nil, fmt.Errorf("retrieving go command flags: %w", err)
+	output := make(map[string]string, len(items))
+	for _, item := range items {
+		if item.Standard && item.ImportPath == "unsafe" && item.Export == "" {
+			// Special-casing "unsafe", because it's not provided like other modules
+			continue
 		}
```
eliottness, line 65: Is `unsafe` really the only module to be like that? What about `import "C"`? There is probably also the `builtins` module on top of my head. Can you extract this logic to make it standout from the rest. Like creating a `virtualPackagesNames` const array or something like this?

main.go
```diff
@@ -183,6 +151,67 @@ func main() {
...
+		buildArgs = append(buildArgs,
+			// All packages we may be instrumenting, plus the standard library.
+			"github.com/datadog/orchestrion/instrument",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/database/sql",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/gin-gonic/gin",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/go-chi/chi.v5",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/gofiber/fiber.v2",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/gorilla/mux",
+			"gopkg.in/DataDog/dd-trace-go.v1/contrib/labstack/echo.v4",
```
eliottness, line 202: Is it really necessary to build the contrib ahead of time even the ones that customers are not using? a standard go build will build the library they are using even without orchestrion, the only new package it the dd-trace-go contrib, not all it's underlying dependencies.

Side question: is a hardcoded list the best way for us to list all the packages we need to build ?

main.go
```diff
@@ -183,6 +151,67 @@ func main() {
...
+
+		log.Tracef("Initializing warm-up module in %q\n", tmp)
+		var stderr bytes.Buffer
+		cmd := exec.Command("go", "mod", "init", "orchestrion-warmup")
+		cmd.Dir = tmp
+		cmd.Stderr = &stderr
+		if err := cmd.Run(); err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to initialize temporary module: %v\n", err)
+			if stderr.Len() > 0 {
+				fmt.Fprintf(os.Stderr, "Error output:\n%s\n", stderr.String())
```
eliottness, line 171: Can you show the full command that was ran somewhere in the output for people to run it manually ?

main.go
```diff
@@ -183,6 +151,67 @@ func main() {
...
+
+		log.Tracef("Running 'orchestrion pin' in the temporary module...\n")
+		stderr.Reset()
+		cmd = exec.Command(orchestrionBinPath, "pin")
+		cmd.Dir = tmp
+		cmd.Stderr = &stderr
+		if err := cmd.Run(); err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to install orchestrion in temporary module: %v\n", err)
+			if stderr.Len() > 0 {
+				fmt.Fprintf(os.Stderr, "Error output:\n%s\n", stderr.String())
```
eliottness, line 184: same here

main.go
```diff
@@ -183,6 +151,67 @@ func main() {
...
+			"gopkg.in/DataDog/dd-trace-go.v1/ddtrace",
+			"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer",
+			"std",
+		)
+		cmd = exec.Command(orchestrionBinPath, buildArgs...)
+		cmd.Dir = tmp
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			fmt.Fprintf(os.Stderr, "Warm-up build failed: %v\n", err)
```
eliottness, line 212: same here

internal/toolexec/aspect/resolve.go
RomainMuller: IIRC this is because I ended up hitting a bug in there during the other stuff... But you're right ^^

internal/toolexec/aspect/resolve.go
RomainMuller: The original `if` branch of that `else` shadows some `err` with its own, and removing the `else` actually changes the meaning of this condition (👏🏻 thank you go, you never disappoint 👏🏻 )

main.go
RomainMuller: This is `warmup` and it's intended largely to platform teams who are providing "ready-made" Docker images... In this case, you want builds made with the image to be fast, and it's okay/expected for the image build itself to pay the toll.

Regarding the ahrd-coded-ness of the list... We need to `//go:generate` this, yes.

internal/toolexec/aspect/resolve.go
eliottness: I would have liked for you to still remove it because the fact that you had had to explain me that makes this code even more confusing. But since this is the only change you would have to make, I will not die on this hill

eliottness (approved):

PR #100, opened 2024-05-23
chore: allow sub-directories for built-in config

RomainMuller:
Allow organizing built-in configuration files with one level of sub-directories, so it's a little easier to manage.


internal/injector/builtin/generator/main.go
```diff
@@ -149,3 +149,38 @@ func writeAll(w io.Writer, data []byte) {
 func init() {
 	log.SetPrefix("github.com/DataDog/orchestrion/internal/injector/builtin/generate: ")
 }
+
+func (g *globs) Set(value string) error {
+	*g = append(*g, value)
+	return nil
```
Hellzy, line 155: Where is this used?

Hellzy (approved):

internal/injector/builtin/generator/main.go
RomainMuller: It's implementing the `flag.Value` interface.

PR #102, opened 2024-05-23
chore: clean up old leftovers

RomainMuller:
Removing a bunch of injected `//dd:` comments that actually don't add any value, and generating a set of blank import statements into the built-in configurations file, so that we no longer have to manually manage the list of packages that need tracking in `go.mod`.


Hellzy (approved):
LGTM overall.
Is there no need to check for duplicates in the various `ImpliesImported` implementations? Granted the behaviour will be correct either way but maybe it will save some iterations/memory later on? You'll know better so up to you

RomainMuller:
> LGTM overall. Is there no need to check for duplicates in the various `ImpliesImported` implementations? Granted the behaviour will be correct either way but maybe it will save some iterations/memory later on? You'll know better so up to you

Deduplication is done at the end of the chain... it reduces complexity quite significantly to do it like that, but we probably wanna keep an eye on this as the complexity grows.

PR #103, opened 2024-05-24
feat: instrument go-redis clients (v7, v8)

RomainMuller:
Add instrumentation logic for `go-redis/redis` clients (v7 and v8).


internal/injector/injector.go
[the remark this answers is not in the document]
RomainMuller: Found a bug & fixed it locally... I guess I could extract that underneath this PR.

.github/workflows/generate.yml
```diff
@@ -24,6 +24,8 @@ jobs:
           cache-dependency-path: '**/*.sum'
       - name: Run 'go generate ./...'
         run: go generate ./...
+      - name: Run 'go mod tidy'
+        run: find . -iname go.mod | xargs dirname | xargs -I DIR go -C DIR mod tidy
```
eliottness, line 28: Does this work: `find -name go.mod -execdir go mod tidy \;`?
it's the same we have on dd-trace-go

_integration-tests/tests/go-redis.v7/validation.json
```diff
@@ -0,0 +1,33 @@
...
+          "resource": "get",
+          "type": "redis",
+          "meta": {
+            "redis.args_length": "2",
+            "component": "go-redis/redis.v7",
+            "out.db": "0",
+            "span.kind": "client",
+            "db.system": "redis",
+            "redis.raw_command": "get test_key: ",
+            "out.host": "localhost"
```
eliottness, line 27: Will this work even with the NAT driver used on windows?

eliottness (approved):
Approved except for the Powershell :stuck_out_tongue:

_integration-tests/tests/go-redis.v7/validation.json
eliottness: I can try this at home if you need

.github/workflows/generate.yml
RomainMuller: Duh I was not aware of `-execdir` (I haven't used this in a while, to be fair)

PR #104, opened 2024-05-28
chore: remove `assignment-of` join-point

RomainMuller:
It is not ideal, as there are many ways to use the result of an expression without necessarily assigning it, which would result in missed instrumentation points. Instead, replace expressions with IIFEs, which is a lot safer.


internal/injector/builtin/yaml/chi.yml
eliottness, line 1: I would still advice to make git properly understand that the `chi.yml` file was renamed and not deleted and recreated, for future all the other PRs waiting to be reviewed that will be rebased on top of it.

eliottness (approved):

PR #105, opened 2024-05-28
feat: support instrumentation of gomodule/redigo

RomainMuller:
Uses a new drop-in replacement primitive, which allows the modified call to stand-in as the original one in stack traces (not showing as generated, which could be confusing).

JIRA: APPSEC-53280


internal/injector/aspect/advice/call.go
```diff
@@ -120,6 +121,53 @@ func (a *appendArgs) AddedImports() []string {
 	return imports
 }
 
+type replaceFunction struct {
+	path string
+	name string
+}
+
+// ReplaceFunction replaces the called function with the provided drop-in replacement. The signature
+// must be compatible with the original function (it may accept a new variadic argument).
```
eliottness, line 130: Is there a way to check if it truely is compatible ? Maybe not at runtime but when generating go code from yaml

.vscode/launch.json
eliottness, line 1: I am positive this should not be in this PR

internal/injector/aspect/advice/call.go
RomainMuller: The issue is that'd need us to correlate the advice with the join-point, which would be non-trivial...

internal/injector/aspect/advice/call.go
```diff
@@ -138,4 +186,23 @@ func init() {
...
 	}
+	unmarshalers["replace-function"] = func(node *yaml.Node) (Advice, error) {
+		var fqn string
+		if err := node.Decode(&fqn); err != nil {
+			return nil, err
+		}
+
+		var (
+			path string
+			name string
```
Hellzy, line 197: That's peanuts but you could declare those together with `fqn`

Hellzy (approved):

PR #107, opened 2024-05-29
chore: use `testcontainers` instead of `setup.ps1`

RomainMuller:
Simplifies the whole management process...


_integration-tests/tests/go-redis.v7/validation.json
```diff
@@ -24,8 +24,7 @@
             "span.kind": "client",
             "db.system": "redis",
             "redis.raw_command": "get test_key: ",
-            "out.host": "localhost",
-            "out.port": "6379"
+            "out.host": "localhost"
```
eliottness, line 27: Is it still needed? Testcontainers returns the address/port pair already

_integration-tests/tests/go-redis.v7/validation.json
```diff
@@ -24,8 +24,7 @@
             "span.kind": "client",
             "db.system": "redis",
             "redis.raw_command": "get test_key: ",
```
eliottness, line 26: Still used?

_integration-tests/tests/go-redis.v8/main.go
```diff
@@ -10,33 +10,68 @@ import (
...
 
 func main() {
+	if os.Getenv("DOCKER_NOT_AVAILABLE") != "" {
+		log.Println("Docker is required to run this test. Exiting with status code 42!")
+		os.Exit(42)
+	}
+
+	ctx := context.Background()
+	server, err := testredis.RunContainer(ctx,
+		testcontainers.WithImage("redis:7"),
```
eliottness, line 33: It is not the redis 8 test?

eliottness (approved):
With how similar the tests between redis 7 and 8 are I am wondering if we could use actual go test cases and run go test and run them in a forloop

_integration-tests/tests/go-redis.v7/validation.json
RomainMuller: This is making sure the correct tags are set on the trace... And I can't assert on the port because it's ephemeral.

_integration-tests/tests/go-redis.v7/validation.json
RomainMuller: This is making sure the correct tags are set on the trace...

_integration-tests/tests/go-redis.v8/main.go
RomainMuller: The _driver_ is v8, but Redis v7 is the latest ^^

PR #108, opened 2024-05-31
feat: instrument net/http clients

RomainMuller:
Instrument `net/http` clients.

JIRA: APPSEC-53473


samples/client/http.go
```diff
@@ -0,0 +1,66 @@
...
+	resp.Body.Close()
+
+	resp, err = http.PostForm("http://localhost:8080", url.Values{"key": {"value"}})
+	if err != nil {
+		log.Fatal(err)
+	}
+	resp.Body.Close()
+}
+
+func shortHandsWithRequest(_ *http.Request /* for context */) {
```
eliottness, line 42: Wait, orchestrion is capable is renaming it and using the generated name in generated and make sure the name is not in conflict with anything ? That's wild.

eliottness: So will the like be considered generated ? Or do you put only the variable name on a line alone?

internal/injector/aspect/advice/import.go
```diff
@@ -0,0 +1,61 @@
...
+	if !hasFile {
+		return false, errors.New("cannot add import: no *dst.File found in node.Chain")
+	}
+
+	refMap, hasMap := typed.ContextValue[*typed.ReferenceMap](ctx)
+	if !hasMap {
+		return false, errors.New("cannot add import: no *typed.ReferenceMap found in context")
+	}
+
+	fmt.Println("Adding import of", a)
```
eliottness, line 40: It's a debug print right ?

internal/injector/aspect/join/method.go
```diff
@@ -0,0 +1,72 @@
...
+	if !ok || node.Recv == nil || node.Recv.NumFields() != 1 || node.Name.Name != p.name {
+		return false
+	}
+
+	if !p.receiver.Matches(node.Recv.List[0].Type) {
+		return false
+	}
+
+	return false
+}
```
eliottness, line 42: So this function always return false ?

internal/injector/aspect/join/method.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	node, ok := chain.Node.(*dst.FuncDecl)
+	if !ok || node.Recv == nil || node.Recv.NumFields() != 1 || node.Name.Name != p.name {
+		return false
+	}
+
+	if !p.receiver.Matches(node.Recv.List[0].Type) {
+		return false
+	}
```
eliottness, line 39:
```suggestion
	if !ok || node.Recv == nil || node.Recv.NumFields() != 1 || node.Name.Name != p.name || !p.receiver.Matches(node.Recv.List[0].Type) {
		return false
	}
```

internal/injector/aspect/join/method.go
eliottness: Since your tests are passing you are definitively missing at least one test

internal/injector/aspect/join/reference.go
```diff
@@ -0,0 +1,62 @@
...
+	return []string{r.importPath}
+
+}
+
+func (r *reference) AsCode() jen.Code {
+	return jen.Qual(pkgPath, "Reference").Call(jen.Lit(r.importPath), jen.Lit(r.name))
+}
+
+func init() {
+	unmarshalers["reference"] = func(node *yaml.Node) (Point, error) {
```
eliottness, line 48: I don't see it used anywhere. Is it supposed to be in this PR?

internal/injector/aspect/join/struct.go
```diff
@@ -78,6 +79,30 @@ func (s *structLiteral) Matches(chain *node.Chain) bool {
...
+				continue
+			}
+			if key.Name == s.field[1:] {
+				return false
+			}
+		}
+
+		return true
+	}
+
```
eliottness, line 105: I also don't see this code used anywhere on the net/http client yml file. Am I missing something?

internal/injector/builtin/yaml/stdlib/net-http.client.yml
eliottness, line 1: I guess it does not hurt to add a comment with a permalink referencing where you took this code

eliottness: Other than that the template code is surprisingly easy to understand with the helper functions

internal/injector/builtin/yaml/stdlib/net-http.client.yml
```diff
@@ -0,0 +1,120 @@
...
+            tracer.Tag(ext.HTTPURL, url.String()),
+            tracer.Tag(ext.Component, "net/http"),
+            tracer.Tag(ext.SpanKind, ext.SpanKindClient),
+            tracer.Tag(ext.NetworkDestinationName, url.Hostname()),
+          }
+          if analyticsRate := globalconfig.AnalyticsRate(); !math.IsNaN(analyticsRate) {
+            opts = append(opts, tracer.Tag(ext.EventSampleRate, analyticsRate))
+          }
+          if port, err := strconv.Atoi(url.Port()); err == nil {
+            opts = append(opts, tracer.Tag(ext.NetworkDestinationPoirt, port))
```
eliottness, line 48:
```suggestion
            opts = append(opts, tracer.Tag(ext.NetworkDestinationPort, port))
```

eliottness (changes requested):

samples/client/http.go
RomainMuller: The variable name will be in generated context; but this won't ever show up in a stack trace anyway...

But yes, it generates a name that is most likely not to cause a conflict (due to the abundance of `_`s in it)... Obviously we can be even smarter in the future, but the complexity of that didn't seem worth the effort at this point...

internal/injector/aspect/join/method.go
RomainMuller: Good catch! And now I wonder how the integration test suite manages to pass... o_O

RomainMuller: Okay the false positive was caused by a leftover `func init() {}` that was hiding the fact the instrumentation was not matching (had it matched, the test would have failed due to double-instrumentation)

.github/workflows/tests.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: ["**"]
   merge_group:
-    branches: ["main"]
+    branches: ['main']
```
eliottness, line 6: Why? I see other double quotes around? Let's reduce the diff

_integration-tests/tests/net_http/validation.json
eliottness, line 1: Nit: a json formatter in a pre-commit could help with this kind of non-mergeable diff

eliottness (approved):
LGTM

.github/workflows/tests.yml
RomainMuller: Right this is copy-pasta from trying to understand why some stuff wasn't working (spoilers: not because of this ^^)

PR #111, opened 2024-06-07
fix: gorilla instrumentation changes Router type

RomainMuller:
This adds a temporary workaround so that `mux.Router` instrumentation does not risk causing compilation errors by changing the value type of the router.


internal/injector/aspect/advice/code/dot_function.go
```diff
@@ -14,6 +14,9 @@ import (
 
 type (
 	function interface {
+		// Receiver returns the name of the receiver of this methid. Fails if the current function is
```
eliottness, line 17:
```suggestion
		// Receiver returns the name of the receiver of this method. Fails if the current function is
```

internal/injector/aspect/advice/code/template.go
```diff
@@ -33,7 +34,18 @@ type Template struct {
 var wrapper = template.Must(template.New("code.Template").Funcs(template.FuncMap{
 	"Version": func() string { return version.Tag },
 }).Parse(
-	"{{define `_`}}package _\nfunc _() {\n{{template `code.Template` .}}\n}{{end}}",
+	`
+{{- define "_statements_" -}}
+package _
+func _() {
+	{{template "code.Template" .}}
```
eliottness, line 41:
```suggestion
	{{ template "code.Template" . }}
```

internal/injector/aspect/advice/code/template.go
```diff
@@ -117,18 +144,22 @@ func (t *Template) compile(ctx context.Context, chain *node.Chain) ([]dst.Stmt,
...
+		}
+		decls = append(decls, dot.placeholders.replaceAllIn(decl).(dst.Decl))
+	}
 
-	list := body.List
-	stmts := make([]dst.Stmt, len(list))
-	for i, node := range list {
-		stmts[i] = t.processImports(ctx, ctxFile, node).(dst.Stmt)
+	for i := range decls {
+		decls[i] = t.processImports(ctx, ctxFile, decls[i]).(dst.Decl)
```
eliottness, line 159: This looks like this is the only usage of `processImports`. Maybe making it return `dst.Decl` would make more sense than casting the result each time

eliottness (commented):
I don't have the time to finish this today but I started at least

eliottness (approved):

PR #115, opened 2024-06-11
refactor: replace inject-source-file with inject-declarations

RomainMuller:
After discovering it is extremely painful to track down & properly handle porting imports from the injected source file into the modified file; and realizing we have full featured support in the `code.Template` object, replaced the `inject-source-file` advice with a new `inject-declarations` version, which accepts a template, which is strictly more flexible.

It also supports specifying a list of `links`, which is packages that need to be linked against to resolve any `//go:linkname` directive within the injected declarations; and specifying any import path in there implies `unsafe` needs to be imported.

### DIFF
--- a/internal/injector/aspect/advice/code/template.go
+++ b/internal/injector/aspect/advice/code/template.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"go/token"
 	"sort"
 	"strconv"
 	"strings"
@@ -33,7 +34,18 @@ type Template struct {
 var wrapper = template.Must(template.New("code.Template").Funcs(template.FuncMap{
 	"Version": func() string { return version.Tag },
 }).Parse(
-	"{{define `_`}}package _\nfunc _() {\n{{template `code.Template` .}}\n}{{end}}",
+	`
+{{- define "_statements_" -}}
+package _
+func _() {
+	{{template "code.Template" .}}
+}
+{{- end -}}
+{{- define "_declarations_" -}}
+package _
+{{ template "code.Template" . }}
+{{- end -}}}
+	`,
 ))
 
 // NewTemplate creates a new Template using the provided template string and
@@ -67,6 +79,12 @@ func (t *Template) CompileBlock(ctx context.Context, node *node.Chain) (*dst.Blo
 	return &dst.BlockStmt{List: stmts}, nil
 }
 
+// CompileDeclarations generates new source based on this Template and extracts
+// all produced declarations.
+func (t *Template) CompileDeclarations(ctx context.Context, node *node.Chain) ([]dst.Decl, error) {
+	return t.compileTemplate(ctx, "_declarations_", node)
+}
+
 // CompileExpression generates new source based on this Template and extracts
 // the produced dst.Expr node. The provided context.Context and *dstutil.Cursor
 // are used to supply context information to the template functions. The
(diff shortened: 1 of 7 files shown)


internal/injector/aspect/advice/code/template.go
```diff
@@ -98,6 +116,15 @@ func (t *Template) CompileExpression(ctx context.Context, node *node.Chain) (dst
 // compile generates new source based on this Template and returns a cloned
 // version of minimally post-processed dst.Stmt nodes this produced.
 func (t *Template) compile(ctx context.Context, chain *node.Chain) ([]dst.Stmt, error) {
+	decls, err := t.compileTemplate(ctx, "_statements_", chain)
+	if err != nil {
+		return nil, err
+	}
+
+	return decls[0].(*dst.FuncDecl).Body.List, nil
```
eliottness, line 124: Is this safe ?

internal/injector/aspect/advice/code/template.go
```diff
@@ -117,18 +144,22 @@ func (t *Template) compile(ctx context.Context, chain *node.Chain) ([]dst.Stmt,
...
+		}
+		decls = append(decls, dot.placeholders.replaceAllIn(decl).(dst.Decl))
+	}
 
-	list := body.List
-	stmts := make([]dst.Stmt, len(list))
-	for i, node := range list {
-		stmts[i] = t.processImports(ctx, ctxFile, node).(dst.Stmt)
+	for i := range decls {
+		decls[i] = t.processImports(ctx, ctxFile, decls[i]).(dst.Decl)
```
eliottness, line 159: `processImports` could definitively return a `dst.Decl` instead of a `dst.Node` to avoid a potentially unsafe case from the futur

internal/injector/aspect/advice/code/template.go
```diff
@@ -33,7 +34,18 @@ type Template struct {
 var wrapper = template.Must(template.New("code.Template").Funcs(template.FuncMap{
 	"Version": func() string { return version.Tag },
 }).Parse(
-	"{{define `_`}}package _\nfunc _() {\n{{template `code.Template` .}}\n}{{end}}",
+	`
+{{- define "_statements_" -}}
+package _
+func _() {
+	{{template "code.Template" .}}
```
eliottness, line 41:
```suggestion
	{{ template "code.Template" . }}
```

eliottness (changes requested):
One last question: Do we need to add a `(*dot).FindImport` method for templates to reference new and existing imports that could have been templated?

internal/injector/aspect/advice/code/template.go
RomainMuller: We control the template, so we know exactly what it can produce at the top-level... So... yeah ^^

internal/injector/aspect/advice/code/template.go
RomainMuller: `processImports` returns the node type it receives (aspirationally); but since it's a method we can't use generics to make it apparent.

eliottness (approved):

RomainMuller:
> One last question: Do we need to add a `(*dot).FindImport` method for templates to reference new and existing imports that could have been templated?

Discussed offline - `dave/dst` handles that for us so we don't need to actually care. Identifier nodes are import-path qualified and DST generates aliases so that they do not collide with local declarations...

PR #117, opened 2024-06-13
fix: linker may fail to satisfy linkdeps unless imported by main

RomainMuller:
Go's `linker` does not bring in all objects from the `importcfg` file unconditionally, instead it only loads objects (transitively) **imported** by the `main` module. This means that _linkDeps_ introduced to satisfy `//go:linkname` directives may not be satisfied by the linker (resulting in undefined symbol errors) unless those dependencies are imported.

This change adds an `OnCompileMain` transformer to the `Weaver` that adds a new synthetic source file that contains blank imports of any yet-to-be-satisfied _linkDeps_ entry, ensuring the linker later satisfies them. It also has the benefit of making sure the compiler correctly registers any `init` function from these dependencies on the start-up stack.

This ought to be safe to do in `main`, since this cannot possibly introduce circular dependencies (and is generally considered good practice if you're using `//go:linkname` anyway).

---

NB - This was discovered as it broke #108, which will hence contribute to creating testing coverage for this change.

### DIFF
--- a/internal/toolexec/aspect/oncompile-main.go
+++ b/internal/toolexec/aspect/oncompile-main.go
@@ -1,0 +1,110 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package aspect
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/datadog/orchestrion/internal/log"
+	"github.com/datadog/orchestrion/internal/toolexec/aspect/linkdeps"
+	"github.com/datadog/orchestrion/internal/toolexec/importcfg"
+	"github.com/datadog/orchestrion/internal/toolexec/proxy"
+	"github.com/dave/jennifer/jen"
+)
+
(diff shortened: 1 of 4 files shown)


internal/toolexec/aspect/oncompile-main.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/datadog/orchestrion/internal/toolexec/aspect/linkdeps"
+	"github.com/datadog/orchestrion/internal/toolexec/importcfg"
+	"github.com/datadog/orchestrion/internal/toolexec/proxy"
+	"github.com/dave/jennifer/jen"
+)
+
+// OnCompileMain only performs changes when compiling the "main" package, adding blank imports for
+// any linkdeps dependencies that are not yet satisfied by the importcfg file. This ensures that the
+// relevant packages' `init` (if any) are appropriately run, and that the linker automatically picks
+// up these dependencies when creating the full binary.
```
Julio-Guerra, line 24: I don't fully agree with this part of the comment: up until now our instrumentations were properly importing the packages they required themselves. But now that you leveraged go:linkname alone, indeed you no longer import the package defining it. So I'd write the rational in this comment around this problem instead.

Julio-Guerra (approved):
Left a comment [about a comment]

PR #124, opened 2024-06-24
chore: add unit test for autoPinOrchestrion / pinOrchestrion

RomainMuller:


pin_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	if _, err := fmt.Fprintln(goMod, "module github.com/datadog/orchestrion/pin-test"); err != nil {
+		return err
+	}
+	if _, err := fmt.Fprintln(goMod); err != nil {
+		return err
+	}
+	if _, err := fmt.Fprintf(goMod, "go %s", runtime.Version()[2:6]); err != nil {
+		return err
+	}
```
Julio-Guerra, line 55: why not just calling go mod init here?

pin_test.go
```diff
@@ -0,0 +1,58 @@
...
+	requiredVersionError = errors.New("test")
+	autoPinOrchestrion()
+	require.NoError(t, requiredVersionError)
+
+	require.FileExists(t, filepath.Join(tmp, orchestrionToolGo))
+	require.FileExists(t, filepath.Join(tmp, "go.sum"))
+
+	data, err := os.ReadFile(filepath.Join(tmp, "go.mod"))
+	require.NoError(t, err)
+	require.Contains(t, string(data), fmt.Sprintf(`github.com/datadog/orchestrion %s`, version.Tag))
```
Julio-Guerra, line 37: `go list` might be a good double-validation that the go.mod file isn't broken

pin_test.go
RomainMuller: It's annoying to shell out I guess...

pin_test.go
RomainMuller: That's not a bad idea :)

eliottness (approved):

PR #125, opened 2024-06-26
fix: missing `link.deps` records cause link-time error

RomainMuller:
In cases when an instrumented dependency is only present in the final binary due to injected dependencies, the `link.deps` file might not contain all necessary information for the linker phase to succeed.

This change adds the necessary logic to properly collect `link.deps` into all archives so they contain the entire closure of link-time dependencies.

Also adds an integration test to validate this is correctly working, and fixes the bug where Orchestrion was not properly loading test files for decoration.


internal/injector/injector.go
```diff
@@ -87,44 +91,64 @@ func New(pkgDir string, opts Options) (*Injector, error) {
...
-				continue
+
+		for _, pkg := range pkgs {
+			switch len(pkg.Errors) {
+			case 0:
+				// Nothing to do, this is a success!
+			case 1:
+				return nil, pkg.Errors[0]
+			default:
+				return nil, fmt.Errorf("%w (and %d more)", pkg.Errors[0], len(pkg.Errors)-1)
```
eliottness, line 113:
```suggestion
				return nil, fmt.Errorf("%d errors occured: %w", len(pkg.Errors), errors.Join(pkg.Errors...))
```

internal/toolexec/aspect/linkdeps/linkdeps.go
```diff
@@ -26,6 +30,22 @@ type LinkDeps struct {
...
+	if err != nil {
+		return res, fmt.Errorf("reading %s from %s[%s]: %w", LinkDepsFilename, importPath, archive, err)
+	}
+	if data == nil {
+		return
+	}
+
+	log.Tracef("Found %s file in %s[%s]\n", LinkDepsFilename, importPath, archive)
+	return Read(data)
+}
```
eliottness, line 47: So `importPath` is only used to print stuff? Can we remove this from the parameters?

internal/injector/aspect/join/directive.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/datadog/orchestrion/internal/injector/node"
+	"github.com/dave/dst"
+	"github.com/dave/jennifer/jen"
+	"gopkg.in/yaml.v3"
+)
+
+type hasDirective string
```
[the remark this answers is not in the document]
RomainMuller: Yeah so I guess we could remove the FunctionOption variant and use `allOf` with the `HasDirective` variant instead...

internal/injector/injector.go
```diff
@@ -49,6 +50,8 @@ type (
 		// Dir is the working directory to use for resolving dependencies, etc... If blank, the current working directory is
 		// used.
 		Dir string
+		// IncludeTests requests test files to be prepared for injection, too.
+		IncludeTests bool
```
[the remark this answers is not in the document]
RomainMuller: I guess we _could_ decide that Orchestrion does not, ever, modify test files, but this feels somewhat arbitrary... If we don't do that, then we have to instead make it so that the injector early-returns if there's _only_ `_test.go` files in the source files list... I'm happy either way (this one way appears to be more versatile?)

internal/injector/injector.go
RomainMuller: Oh you spotted a leftover form before the "go1.21 or bust" era :)

internal/toolexec/aspect/linkdeps/linkdeps.go
RomainMuller: The logs are much more difficult to process without it...

internal/injector/aspect/join/directive.go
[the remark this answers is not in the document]
RomainMuller: - `valueDeclaration` could have more fields in the future (name pattern, etc...)
- `hasDirective` checks for the presence of a directive with a given name, and has little chance to ever support more

internal/toolexec/aspect/linkdeps/linkdeps.go
eliottness: `errors.Join` is go1.20 right?

eliottness: The diff tells me that this was in a fmt.Errorf higher in the stack before, let's put it where it came from and reduce the context to take to read this function

internal/injector/aspect/advice/assign.go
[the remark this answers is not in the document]
RomainMuller: Good catch :) Found that I did not use `go tool covdata` properly and it was hence not using the new test's coverage when producing the report... >_>

eliottness (approved):

PR #126, opened 2024-06-27
chore: replace PowerShell integration test with go test suite

RomainMuller:
Replaces the PowerShell script (`integration-tests.ps1`) with a new Go-native test suite that is to be run orchestrion-instrumented. This considerably simplifies the overall test structure.


_integration-tests/utils/agent/agent.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	venvBin := filepath.Join(agent.virtualEnv, "bin")
+	if runtime.GOOS == "windows" {
+		venvBin = filepath.Join(agent.virtualEnv, "Scripts")
+	}
+
+	t.Logf("Installing ddapm-test-agent in venv...\n")
+	if err = exec.Command(filepath.Join(venvBin, "pip"), "install", "ddapm-test-agent").Run(); err != nil {
+		return
+	}
```
eliottness, line 56: Requests Changes: I believe it makes more sense here to create a `.venv` folder in the current repository. We can then add it to the .gitignore and make it part of the CI cache.

Nit: We could also make this code part of simple `integration-tests.(sh|ps1)` ( :stuck_out_tongue_winking_eye: ) at the root that would run the test-agent and `go test -build-tags=integration`

_integration-tests/utils/agent/agent.go
```diff
@@ -0,0 +1,188 @@
...
+	currentSession atomic.Pointer[mockSession]
+	port           int
+}
+
+type mockSession struct {
+	agent *MockAgent
+	token *guid.Guid
+}
+
+func New(t *testing.T) (agent MockAgent, err error) {
```
eliottness, line 40: This is not necessary to use named return here because there is no `defer` code so please don't use it as it makes the flow very weird as always

_integration-tests/utils/agent/agent.go
```diff
@@ -0,0 +1,188 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/beevik/guid"
```
eliottness, line 23: `github.com/google/uuid` is already in the dependency list because of dd-trace-go so let's use it instead of adding something new to the go.mod

_integration-tests/tests/suite.generated.go
```diff
@@ -0,0 +1,42 @@
...
+	"go-redis.v7":     new(goredisv7.TestCase),
+	"go-redis.v8":     new(goredisv8.TestCase),
+	"gorm":            new(gorm.TestCase),
+	"gorm.jinzhu":     new(gormjinzhu.TestCase),
+	"grpc":            new(grpc.TestCase),
+	"mux":             new(mux.TestCase),
+	"net_http":        new(nethttp.TestCase),
+	"redigo":          new(redigo.TestCase),
+	"sql":             new(sql.TestCase),
+}
```
eliottness, line 42: when re-running this locally this looks like this was not gofmt'd #longliveprecommit

eliottness (commented):
First look without looking at the tests files themselves

eliottness (commented):
First look without looking at the tests files themselves

_integration-tests/tests/suite.go
```diff
@@ -0,0 +1,22 @@
...
+
+//go:generate go run ../utils/generator
+
+type testCase interface {
+	Setup(*testing.T)
+	Run(*testing.T)
+	Teardown(*testing.T)
+
+	ExpectedTraces() trace.Spans
+}
```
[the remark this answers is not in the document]
eliottness: Suggestion: Add a small doc on integration tests

_integration-tests/tests/suite_test.go
```diff
@@ -0,0 +1,61 @@
...
+func Test(t *testing.T) {
+	require.True(t, orchestrionEnabled, "this test suite must be run with orchestrion enabled")
+	require.NotEmpty(t, suite, "no test case registered")
+
+	mockAgent, err := agent.New(t)
+	require.NoError(t, err)
+	defer mockAgent.Close()
+
+	for name, tc := range suite {
+		tc := tc
```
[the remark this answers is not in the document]
RomainMuller: That wouldn't work for testing with `go1.21`

_integration-tests/utils/agent/agent.go
RomainMuller: >Suggestion: We could also make this code part of simple `integration-tests.(sh|ps1)` ( 😜 ) at the root that would run the test-agent and go test `-build-tags=integration`

I specifically want to avoid having folks use anything other than `go` or `orchestrion go` commands, so I specifically want to NOT have such a script anymore.

_integration-tests/utils/agent/agent.go
RomainMuller: I don't see what difference it makes, but okay :)

_integration-tests/tests/suite_test.go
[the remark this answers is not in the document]
eliottness: What is limiting us from moving up to go 1.22?

_integration-tests/utils/agent/agent.go
RomainMuller: I've made it possible to use a "from-env" binary, and the CI tests cache that now...

_integration-tests/tests/grpc/main.go
eliottness, line 1: `validation.json` is still here

eliottness (changes requested):

_integration-tests/tests/go-redis.v7/main.go
```diff
@@ -3,43 +3,39 @@
...
+
+	var err error
+	tc.server, err = testredis.RunContainer(ctx,
+		testcontainers.WithLogger(testcontainers.TestLogger(t)),
 		testcontainers.WithImage("redis:7"),
-		testcontainers.WithLogConsumers(&testcontainers.StdoutLogConsumer{}),
-		testcontainers.WithHostConfigModifier(func(config *container.HostConfig) {
-			if runtime.GOOS == "windows" {
-				config.NetworkMode = network.NetworkNat
-			}
```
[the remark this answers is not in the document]
RomainMuller: Can't test that as it is impossible to do on GH Actions, so I prefer to leave it out...

_integration-tests/tests/go-redis.v7/main.go
```diff
@@ -49,11 +45,10 @@ func main() {
 		),
 	)
 	if err != nil {
-		log.Fatalf("Failed to start redis test container: %v\n", err)
+		t.Skipf("Failed to start redis test container: %v\n", err)
```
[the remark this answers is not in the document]
RomainMuller: It used to not run until that point when Docker isn't available. Now it does, and we take this as the signal the platform is not supported.

.github/workflows/tests.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: go${{ steps.simple-go-version.outputs.version }},${{ runner.os }},${{ runner.arch }},unit
-          files: ./coverage/unit.out,.couverage/integration.out
+          files: ./coverage/unit.out,.coverage/integration.out
```
eliottness, line 72:
```suggestion
          files: ./coverage/unit.out,./coverage/integration.out
```

eliottness (approved):
LGTM, please fix the last issue and then merge it

PR #127, opened 2024-06-28
fix: cannot lazily link against GLS accessors

RomainMuller:
In order to be able to lazily link against GLS accessors, these need to be declared as variables instead of functions. This allows creation of bindings on the other side that can be conditionally satisffied by the existence of initialized declarations somewhere else.

Adds an integration test to demonstrate it is possible to build programs with and without the runtime modification using this technique.

Enables: DataDog/dd-trace-go#2761

JIRA: APPSEC-53654


internal/injector/builtin/yaml/stdlib/runtime.yml
```diff
@@ -18,18 +18,19 @@ aspects:
   - join-point:
       struct-definition: runtime.g
     advice:
+      - add-blank-import: unsafe
       - add-struct-field:
           name: __dd_gls
           type: any
       - add-blank-import: unsafe # Needed for go:linkname
-      - inject-declarations:
+      - inject-declarations: # Reference: https://github.com/golang/go/blob/master/src/runtime/HACKING.md?plain=1#L44
```
eliottness, line 26: permalink

felixge (approved):

PR #158, opened 2024-07-11
ci: setup reusable workflow for integration tests

eliottness:
### What does this PR do?

This PR creates a new workflow because we can't use the codecov workflow in dd-trace-go. Even if we are not technically using it because the `if` condition would have evaluated to false. cf. [this job](https://github.com/DataDog/dd-trace-go/actions/runs/9892772923)

### Motivation

Add a CI connection dd-trace-go and orchestrion to make sure we don't break orchestrion features when changing dd-trace-go

### DIFF
--- a/.github/workflows/workflow_call.yml
+++ b/.github/workflows/workflow_call.yml
@@ -26,18 +26,13 @@ jobs:
         runs-on: [macos, ubuntu, windows]
         go-version: ['1.21', '1.22']
     runs-on: ${{ matrix.runs-on }}-latest
-    name: Integration tests (go${{ matrix.go-version }}, ${{ matrix.runs-on }}, ${{ matrix.build-mode }})
+    name: Integration tests (go${{ matrix.go-version }}, ${{ matrix.runs-on }})
     steps:
       - name: Checkout orchestrion
         uses: actions/checkout@v4
         with:
+          path: orchestrion
           repository: DataDog/orchestrion
-      - name: Checkout dd-trace-go
-        uses: actions/checkout@v4
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow_call.yml
```diff
@@ -0,0 +1,61 @@
...
+        default: main
+  workflow_call:
+    inputs:
+      dd-trace-go-ref:
+        type: string
+        required: true
+        description: 'The ref to checkout dd-trace-go at'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
RomainMuller, line 18: `github.event.pull_request.number` is never going to be set; because this does not trigger from `pull_request`. And `github.ref` is (I think) going to be the current tip of `main` on this repository... This means multiple PRs from a remote repo would cancel each other.

Probably this would be better:
```suggestion
  group: ${{ github.workflow }}-${{ inputs.dd-trace-go-ref }}
```

.github/workflows/workflow_call.yml
```diff
@@ -0,0 +1,61 @@
...
+  cancel-in-progress: true
+
+jobs:
+  integration-tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        runs-on: [macos, ubuntu, windows]
```
RomainMuller, line 26: I wonder if we _need_ to run on all OS'es there or if we can prune to just `ubuntu` instead... I have no problem testing all three; just the macOS and Windows runners are slow & busy...

RomainMuller (approved):

PR #164, opened 2024-07-15
feat: support instrumentation of `go.mongodb.org/mongo-driver/mongo`

darccio:
### What does this PR do?

Adds support for instrumenting `go.mongodb.org/mongo-driver/mongo`.

### Reviewer's Checklist


internal/injector/builtin/yaml/databases/mongo.yml
```diff
@@ -0,0 +1,25 @@
...
+        - function-call: go.mongodb.org/mongo-driver/mongo.Connect
+    advice:
+      - wrap-expression:
+          imports:
+            redis: go.mongodb.org/mongo-driver/mongo
+            mongotrace: gopkg.in/DataDog/dd-trace-go.v1/contrib/go.mongodb.org/mongo-driver/mongo
+          template: |-
+            func() (client *mongo.Client, err error) {
+              opts.Monitor = mongotrace.NewMonitor()
+              return mongo.Connect(ctx, opts)
```
RomainMuller, line 24: You need to have `mongo` defined in the `imports` map (it needs to bind all package names to the corresponding import path, even if the context implies this package is already imported, because this snippet is parsed in a separate context before being moved into the final updated AST)

internal/injector/builtin/yaml/databases/mongo.yml
```diff
@@ -0,0 +1,25 @@
...
+---
+meta:
+  name: go.mongodb.org/mongo-driver/mongo
+  description: Official Golang driver for MongoDB.
+  icon: database
+aspects:
+  - id: Inject mongotrace.NewMonitor
+    join-point:
+      one-of:
+        - function-call: go.mongodb.org/mongo-driver/mongo.Connect
```
RomainMuller, line 15: No need to `one-of` if there is only one option :)

```suggestion
    join-point:
      function-call: go.mongodb.org/mongo-driver/mongo.Connect
```

internal/injector/builtin/testdata/client/mongo.go.snap
```diff
@@ -0,0 +1,38 @@
...
+func mongoClient() {
+	ctx := context.Background()
+	opts := options.Client().ApplyURI("mongodb://localhost:27017")
+	client, err :=
+//line <generated>:1
+		func() (client *mongo.Client, err error) {
+			opts.Monitor = mongo.NewMonitor()
+			return mongo.Connect(ctx, opts)
+		}()
+//line samples/client/mongo.go:20
```
RomainMuller, line 27: If `opts` is necessarily created via `options.Client()`, then maybe we can instrument around `options.Client()`, adding the `.SetMonitor(mongo.NewMonitor())` call instead of doing the IIFE around the `mongo.Connect` call?

RomainMuller (changes requested):

internal/injector/builtin/yaml/databases/mongo.yml
darccio: @RomainMuller You are right. I left a redis prefix from the example I based this PR.

internal/injector/builtin/testdata/client/mongo.go.snap
darccio: @RomainMuller Done, I think it's the best way. I tried it previously but it wasn't working, but now I saw why: I was using `mongo.Client` instead of `options.Client`.

internal/injector/builtin/yaml/databases/mongo.yml
```diff
@@ -0,0 +1,25 @@
...
+    advice:
+      - wrap-expression:
+          imports:
+            redis: go.mongodb.org/mongo-driver/mongo
+            mongotrace: gopkg.in/DataDog/dd-trace-go.v1/contrib/go.mongodb.org/mongo-driver/mongo
+          template: |-
+            func() (client *mongo.Client, err error) {
+              opts.Monitor = mongotrace.NewMonitor()
```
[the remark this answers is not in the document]
darccio: After applying the feedback, instrumenting around `options.Client` there is no need to check if `opts.Monitor` is nil, as it'll be replaced later by the customer code if they assign it or call `SetMonitor`.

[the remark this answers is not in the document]
RomainMuller: That is correct :)

RomainMuller (approved):

PR #175, opened 2024-07-19
WIP: instrument net/http.ServeHTTP implementations

### LINKED ISSUE #173: net/http: `Handler` interface implementation is not always instrumented
**Version of orchestrion**
`0.7.x`

**Describe what happened:**

The implementation of the `ServerHTTP` method does not get instrumented, because its aspect is gated by `httpmode=report`, which cannot be enabled externally.

```go
// You can edit this code!
// Click here and start typing.
package main

import "net/http"

type Handler struct{}

func (Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {}

func main() {
	// NB: http.Handle(...), using the default ServeMux, doesn't get instrumented either
	mu := http.NewServeMux()
	mu.Handle("/foo", Handler{})
	http.ListenAndServe(":8080", mu)
}

```

nsrip-dd:
### What does this PR do?

TODO - tests

Makes sure we instrument `net/http` handlers and servers.

#### Background

There are two general ways to start an HTTP server with `net/http`:

- Package-level functions:
  - `http.ListenAndServe`
  - `http.ListenAndServeTLS`
  - `http.Serve`
  - `http.ServeTLS`
- Via the `net/http.Server` methods:
  - `http.(*Server).ListenAndServe`
  - `http.(*Server).ListenAndServeTLS`
  - `http.(*Server).Serve`
  - `http.(*Server).ServeTLS`
 
The package-level functions create a `Server` and dispatch to the matching methods.

All of these work with the `http.Handler` interface, which defines the "application" logic for handling an HTTP request. The `Server` type has a `Handler` field, and all of the package-level functions take an `http.Handler` argument. Many packages, like `github.com/gorilla/mux`, provide `http.Handler` implementations.

The `net/http` package has two `http.Handler` implementations: `net/http.ServeMux`, and `net/http.HandlerFunc`.

#### What this PR actually does

Before, we instrumented the `Handler` field of `Server` literals, and function literals cast to `http.HandlerFunc`. But this could lead to redundant instrumentation, and also didn't cover common patterns. For example, this doesn't get instrumented today:

```go
package main

import "net/http"

type Handler struct{}

func (Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {}

func main() {
	mu := http.NewServeMux()
	mu.Handle("/foo", Handler{})
	http.ListenAndServe(":8080", mu)
}
```

For another example, we get redundant instrumentation for [this](https://github.com/DataDog/orchestrion/blob/0acfd197a7a60487e92ad646cca9a143fe839aec/_integration-tests/tests/mux/mux.go#L31C1-L42C19) `github.com/gorilla/mux` example:

```go
mux := mux.NewRouter()
tc.Server = &http.Server{
	Addr:    "127.0.0.1:" + utils.GetFreePort(t),
	Handler: mux,
}

mux.HandleFunc("/ping", func(w http.ResponseWriter, _ *http.Request) {
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)
	_, err := io.WriteString(w, `{"message": "pong"}`)
	assert.NoError(t, err)
}).Methods("GET")
```

The `Handler` field of the `Server` literal gets instrumented, _and_ the `http.HandlerFunc` literal is instrumented, _and_ the router returned via `mux.NewRouter` is instrumented. We have at least one too many layers of instrumentation.

So this PR takes this approach:
- We define a new `MaybeWrapHandler` function. It takes an `http.Handler` and returns an `http.Handler`. If the input handler is one of the `net/http` handler types, we wrap it with tracing instrumentation. Otherwise we pass it on unchanged under the assumption it is already instrumented.
- We use this function to _maybe_ wrap the `http.Handler` passed to the package-level functions.
- We also use this function to _maybe_ wrap the `Handler` field of a `Server` literal, similar to what we did before.

This doesn't provide complete coverage. For example, the user could do something like `s := new(http.Server); s.Handler = foo` and we wouldn't catch it. But this hopefully improves our coverage significantly and avoids _some_ redundant instrumentation.

### Motivation

Fixes #173

### Reviewer's Checklist

### DIFF
--- a/_integration-tests/tests/net_http/net_http.go
+++ b/_integration-tests/tests/net_http/net_http.go
@@ -26,6 +26,22 @@ type TestCase struct {
 	*http.Server
 }
 
+type handler struct{}
+
+func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	b, err := io.ReadAll(r.Body)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Write([]byte(err.Error()))
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	w.Write(b)
+}
+
 func (tc *TestCase) Setup(t *testing.T) {
 	mux := http.NewServeMux()
 	tc.Server = &http.Server{
@@ -35,6 +51,21 @@ func (tc *TestCase) Setup(t *testing.T) {
 
 	mux.HandleFunc("/hit", tc.handleHit)
 	mux.HandleFunc("/", tc.handleRoot)
+	// TODO: hit these endpoints, check for spans
+	mux.Handle("/handler", handler{})
+	mux.HandleFunc("/literal", func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+
+		b, err := io.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte(err.Error()))
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		w.Write(b)
+	})
 
 	go func() { assert.ErrorIs(t, tc.Server.ListenAndServe(), http.ErrServerClosed) }()
 }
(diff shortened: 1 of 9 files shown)


nsrip-dd:
Closing in favor of #403

PR #176, opened 2024-07-22
feat: orchestrion job server to reduce duplicated effort

RomainMuller:
A new integrated job-server using the NATS protocol is used to share effort between all involved processes. This reduces the amount of duplicated work that may be going on; while remaining auditable.

The job-server is a simple request-response server. A server is started automatically when relevant, listening to a randomly chosen available port on the loopback interface (via address `127.0.0.1`):

- When using `orchestrion go [build|list|...]`, the job-server is started by the root `orchestrion` process and shared to all children via a dedicated environment variable;
- When using `-toolexec="orchestrion toolexec"` (either directly or via `GOFLAGS`), a new job-server is started by the first `toolexec` invocation that needs one; and the connect URL is stored in a specific file within the `go` toolchain's `$WORK` directory, so that other `toolexec` processes can discover and use it. Such servers automatically shut down once the `$WORK` directory is destroyed, or if they have not had any client connected for a set time (`15` minutes).

Job servers maintain a cache of responses provided to requests, and only compute a given result once. As of this PR, the job server handles:

- Computing the version information added to `-V=full` command invocations, although this can only benefit from a shared job server if the build is started by `orchestrion go`, as otherwise these commands are invoked _before_ the `$WORK` directory is created;
- Resolving injected packages' export files, which is a common source of duplicated effort, as the `go toolchain` did not know about these and hence did not account for these dependencies when building the job tree, such that multiple concurrent resolutions may be attempted.


internal/jobserver/server.go
```diff
@@ -0,0 +1,314 @@
...
+		}
+
+		res.inactivityTimeout = opts.InactivityTimeout
+		res.clients = make(map[uint64]string)
+		if _, err := sysConn.Subscribe(fmt.Sprintf("$SYS.ACCOUNT.%s.CONNECT", userAccount.Name), res.handleClientConnect); err != nil {
+			return nil, err
+		}
+		if _, err := sysConn.Subscribe(fmt.Sprintf("$SYS.ACCOUNT.%s.DISCONNECT", userAccount.Name), res.handleClientDisconnect); err != nil {
+			return nil, err
+		}
```
eliottness, line 179: Are we sure this won't cause client-side issue if the handlers are not even subscribed?

internal/cmd/server.go
```diff
@@ -0,0 +1,149 @@
...
+			Usage: "Automatically shut down after a period without any connected client",
+			Value: time.Minute,
+		},
+		&cli.BoolFlag{
+			Name:  "nats-logging",
+			Usage: "Enable NATS server logging",
+		},
+	},
+	Hidden: true,
+	Action: func(c *cli.Context) error {
```
eliottness, line 49: There is about 70% of this function being behind `urlFile != ""` two if statements. How about reversing the logic here or at least moving it in another function dedicated to the non-happy path ?

internal/cmd/server.go
```diff
@@ -0,0 +1,149 @@
...
+			defer os.Remove(urlFile)
+		}
+
+		// Start the server for real now...
+		server, err := jobserver.New(&opts)
+		if err != nil {
+			return cli.Exit(fmt.Errorf("failed to start NATS server: %w", err), 1)
+		}
+
+		if urlFile != "" {
```
eliottness, line 100: Above suggestion extents to this if statement :point_up:

internal/cmd/server.go
```diff
@@ -0,0 +1,149 @@
...
+				if err := mu.Unlock(); err != nil {
+					return cli.Exit(fmt.Errorf("failed to unlock %q: %w", urlFile, err), 1)
+				}
+			}
+
+			// Watch for removal of the URL file, and shut down the server if/when that happens... Note that if we are unable
+			// to watch, we'll leave it entirely up to the inactivity timeout instead...
+			watcher, err := fsnotify.NewWatcher()
+			if err != nil {
+				log.Warnf("Failed to create file watcher: %v\n", err)
```
eliottness, line 117: This `if` is very inconspicuous but brings great issues. How about explaining the issue clearly in 2/3 lines for the user?

internal/jobserver/server.go
```diff
@@ -0,0 +1,314 @@
...
+	})
+}
+
+// Tries to identify a loopback IP address from available interfaces. This is
+// done to ensure the server will work even if the host runs an IPv6-only
+// stack, as we would discover `::1` appropriately.
+func init() {
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		return
```
eliottness, line 306: Maybe logging something here is not too bad

internal/jobserver/client/env.go
```diff
@@ -0,0 +1,123 @@
...
+//   - If the ORCHESTRION_JOBSERVER_URL environment variable is set, a client
+//     connected to this URL is returned.
+//   - Otherwise, if workDir is not empty, a server will be identified based on
+//     a `.orchestrion-jobserver` file; or a new server will be started using
+//     that url file, and a connection will be established to it. The started
+//     job server will automatically shut itself down once it no longer has any
+//     active client for a period of time.
+//   - Otherwise, the ErrNoServerAvailable error is returned.
+func FromEnvironment(workDir string) (*Client, error) {
+	if client == nil {
```
eliottness, line 42: Can you revert this if statement and make it a returning-if to lower tabulation levers?

internal/jobserver/client/env.go
```diff
@@ -0,0 +1,123 @@
...
+		if url := os.Getenv(ENV_VAR_JOBSERVER_URL); url != "" {
+			log.Debugf("Connecting to job server at %q (from %s)\n", url, ENV_VAR_JOBSERVER_URL)
+			if c, err := Connect(url); err != nil {
+				return nil, err
+			} else {
+				client = c
+			}
+		} else if workDir != "" {
```
eliottness, line 50: Same here with the code line 94-95

eliottness: Or make all of this a switch case as you wish

internal/jobserver/client/env.go
```diff
@@ -0,0 +1,123 @@
...
+					break
+				} else if cmd.ProcessState == nil && time.Since(startTime) <= timeout {
+					log.Tracef("Job server still not ready in %q...\n", urlFilePath)
+					time.Sleep(150 * time.Millisecond)
+				} else {
+					// Attempt to kill the process if it hasn't died by itself...
+					_ = cmd.Process.Kill()
+					return nil, err
+				}
+			}
```
eliottness, line 88: You should be able to put this in a separate function without too much issues

eliottness (changes requested):
Overall looks very good already but very tiring to review

internal/jobserver/client/client.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/datadog/orchestrion/internal/jobserver/common"
+	"github.com/nats-io/nats.go"
+)
+
+const (
+	USERNAME    = "orchestrion"
+	NO_PASSWORD = "" // We only use account management to have access to system events, not for security.
+
+	HDR_CLIENT_PID  = "client-pid"  // Header containing the client PID
+	HDR_CLIENT_PPID = "client-ppid" // Header containing the client parent PID
```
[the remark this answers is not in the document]
RomainMuller: Goland is right! This is left over from an experiment I ended up not liking AT ALL.

internal/jobserver/run.go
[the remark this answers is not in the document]
RomainMuller: Ooops, forgot to delete!

internal/log/level.go
```diff
@@ -37,6 +37,10 @@ func LevelNamed(name string) (Level, bool) {
 	}
 }
 
+func (l Level) Printf(format string, args ...any) {
+	write(l, format, args...)
+}
+
```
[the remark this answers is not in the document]
RomainMuller: Ah this is also dead code from back when I made this experiment I didn't like at all ^^

eliottness (approved):
LGTM

PR #177, opened 2024-07-22
fix: injected imports conflict with local names

### LINKED ISSUE #174: Injected imports can clash with existing package-scope type, variable, const names
**Version of orchestrion**
main

**Describe what happened:**

I built the following program with Orchestrion:

```
package main

import "context"

type event struct {}

//dd:span
func DoThing(ctx context.Context, e event) {}

func main() {
	DoThing(context.Background(), event{})
}
```

I saw the following error:

```
# testapp
./main.go:5: event already declared through import of package event ("github.com/datadog/orchestrion/instrument/event")
	<generated>:2[/var/folders/f3/g91d13pd6kd3vdxts_gsgd1r0000gn/T/go-build1208255770/b001/orchestrion/src/main/main.go:8:2]: other declaration of event
./main.go:8: use of package event not in selector
./main.go:11: use of package event not in selector
```

The program to build successfully, with span instrumentation added to `DoThing`.

There is name mangling to handle conflicting imports (~I haven't found _where_, but I know it happens :)~ looks like [this package](https://pkg.go.dev/github.com/dave/jennifer/jen) does it?). But we can still have clashes between package names and package-scope type/variable/const names. I noticed this when working on #173, where adding `event` imports was clashing with `event` types in other packages. See [this failure](https://github.com/DataDog/orchestrion/actions/runs/10012699280/job/27678748028#step:7:138), for example.

RomainMuller:
It appears that `dave/dst` will emit imports with aliases in such a way to avoid naming collisions between imports, but does not consider the possible collisions with symbols elsewhere in the AST.

In order to address this issue, we eagerly specify aliases for synthetic imports that are designed to be extremely unlikely to conflict with any user-defined symbol in reasonable source code (formatted as `__orchestrion_%s`). `dave/dst` will use such aliases as a base when emitting final imports, and will automatically mangle them as needed if there is any conflict between them.

Fixes #174


nsrip-dd (commented):
Thanks! The changes generally look good to me. A few things:

---

Confirmed the fix works for the example in #174. It would be nice to have a regression test for this issue. Right now none of the tests break if I revert the fix in this PR. How about something like this, modifying the injector test to try compiling the output (optional so we don't have to rework every test case to fully build):

```diff
diff --git a/internal/injector/injector_test.go b/internal/injector/injector_test.go
index 5bc3649..bc26c1e 100644
--- a/internal/injector/injector_test.go
+++ b/internal/injector/injector_test.go
@@ -32,6 +32,7 @@ type testConfig struct {
 	PreserveLineInfo    bool                           `yaml:"preserveLineInfo"`
 	SyntheticReferences map[string]typed.ReferenceKind `yaml:"syntheticReferences"`
 	Code                string                         `yaml:"code"`
+	MustCompile         bool                           `yaml:"mustCompile"`
 }
 
 func Test(t *testing.T) {
@@ -92,6 +93,12 @@ func Test(t *testing.T) {
 			require.NoError(t, err, "failed to read modified file")
 			modified := normalize(modifiedData, inputFile)
 
+			if config.MustCompile {
+				newPath := filepath.Join(tmp, "output.go")
+				os.Rename(res.Filename, newPath)
+				runGo(t, tmp, "build", newPath)
+			}
+
 			assert.Equal(t, config.SyntheticReferences, res.References.Map())
 
 			edits := myers.ComputeEdits(span.URIFromPath("input.go"), original, modified)
diff --git a/internal/injector/testdata/injector/name-conflict/config.yml b/internal/injector/testdata/injector/name-conflict/config.yml
new file mode 100644
index 0000000..a43d3c6
--- /dev/null
+++ b/internal/injector/testdata/injector/name-conflict/config.yml
@@ -0,0 +1,29 @@
+%YAML 1.1
+---
+aspects:
+  - join-point:
+      function-body:
+        all-of:
+          - directive: 'dd:span'
+    advice:
+      - prepend-statements:
+          imports:
+            fmt: fmt
+          template: |-
+            fmt.Println("hello world")
+
+preserveLineInfo: true
+mustCompile: true
+
+syntheticReferences:
+  fmt: true
+
+code: |-
+  package test
+
+  var fmt = "hello"
+
+  //dd:span
+  func foo() {
+    println(fmt)
+  }
diff --git a/internal/injector/testdata/injector/name-conflict/expected.diff b/internal/injector/testdata/injector/name-conflict/expected.diff
new file mode 100644
index 0000000..ea058ab
--- /dev/null
+++ b/internal/injector/testdata/injector/name-conflict/expected.diff
@@ -0,0 +1,20 @@
+--- input.go
++++ output.go
+@@ -1,8 +1,17 @@
+ package test
+ 
++//line <generated>:1
++import __orchestrion_fmt "fmt"
++
++//line input.go:3
+ var fmt = "hello"
+ 
+ //dd:span
+ func foo() {
++//line <generated>:1
++  {
++    __orchestrion_fmt.Println("hello world")
++  }
++//line input.go:7
+   println(fmt)
+ }
```

If I make that change, and drop the fix by making ReferenceMap.AddImport ignore the alias, I get a failure like this:

```
go: creating new go.mod: module dummy/test/module
# command-line-arguments
./input.go:3: fmt already declared through import of package fmt ("fmt")
	<generated>:1[/var/folders/f3/g91d13pd6kd3vdxts_gsgd1r0000gn/T/Testname-conflict4277439739/001/output.go:4:8]: other declaration of fmt
./input.go:7: use of package fmt not in selector
```

And if I restore the fix, the test passes. I'm open to other ideas, of course :)

---

[This failure looks related to these changes](https://github.com/DataDog/orchestrion/actions/runs/10044511172/job/27759724919?pr=177#step:7:137):
```
# github.com/gorilla/mux
<generated>:10: undefined: tracer
```
The failure specifically happens due to [this rule](https://github.com/DataDog/orchestrion/blob/661f36a1c538988bac4d58813085e0e1a43a4820/internal/injector/builtin/yaml/http/gorilla.yml#L107-L147). Specifically, [this line](https://github.com/DataDog/orchestrion/blob/661f36a1c538988bac4d58813085e0e1a43a4820/internal/injector/builtin/yaml/http/gorilla.yml#L125) calls into the tracer library, but it's missing from the group of imports for the rule. I _think_ it coincidentally worked before because another rule already added a tracer import, but now it's broken since the tracer import is aliased? So I think all that's needed is to add a tracer import to that rule.

RomainMuller:
Actually I tried and it's trivial to make all tests' outputs compile, so I went ahead and made a compile sanity test _mandatory_ here, because it feels like the right thing to do.

I've also added your regression test to the mix :)

nsrip-dd (approved):

PR #178, opened 2024-07-22
feat: add continuous profiler instrumentation

nsrip-dd:
PROF-10258

### What does this PR do?

Add profiling, enabled manually by setting `DD_PROFILING_ENABLED=true` at run time.
The profiler has roughly the default configuration, with a few non-default profile types which
should be low-enough overhead and sufficiently useful to enable for recent Go releases.
There are not currently any other configuration knobs.

If enabled this way, the profiles will be tagged with `orchestrion:true`.

This PR also accepts the value `DD_PROFILING_ENABLED=auto` to enable profiling.
This can be provided via the [Datadog Admission Controller](https://docs.datadoghq.com/containers/cluster_agent/admission_controller) (see [this PR](https://github.com/DataDog/datadog-agent/pull/27185)).
Some languages/runtimes use heuristics to decide whether to enable profiling when
`auto` is provided, to avoid profiling short-lived or non-instrumented applications.
We assume the application is meant to be instrumented by virtue of the user building
with Orchestrion. And the Go profiler won't send any data until at least one minute has
passed. So, we treat `auto` the same as `true`.

TODO - document this once it's released

TODO - this has only been manually tested. That's probably okay for now; there's not
much to this code. But we can investigate ways to automatically test it.

### Motivation

If somebody uses Orchestrion to add APM instrumentation, and they also want
profiling, Orchestrion should be able to add it so the user doesn't have to
separately modify their code to get profiling.

### Reviewer's Checklist


internal/injector/builtin/yaml/go-main.yml
```diff
@@ -21,7 +21,41 @@ aspects:
...
+            os: os
+            slices: slices
+            strings: strings
           template: |-
             tracer.Start(tracer.WithOrchestrion(map[string]string{"version": {{printf "%q" Version}}}))
             defer tracer.Stop()
 
+            // TODO: this probably premature, not sure if "autoinjection"
+            // will actually be a thing for Go binaries.
+            injected := slices.Contains(
```
Julio-Guerra, line 34: side note while I pass through this PR: you must enforce a separate lexical scope scope for those statements to avoid name clashes with the rest of the main function we don't know about:

```
{
    // your instrumentation block
}
```

@RomainMuller often uses the [IIFE](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) pattern for this too, but the block should be enough from what I can see here.

nsrip-dd: Good to know, thanks! Looks like it's already put in its own scope: https://github.com/DataDog/orchestrion/pull/178/files#diff-c4820fd0f6a6380e158a2a9cd02a46aa7da2df4879c4f9914ab76075d9f5498aR24-R49

internal/injector/builtin/yaml/go-main.yml
```diff
@@ -21,7 +21,38 @@ aspects:
       - prepend-statements:
           imports:
             tracer: gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
+            profiler: gopkg.in/DataDog/dd-trace-go.v1/profiler
+            log: log
+            os: os
+            slices: slices
+            strings: strings
```
RomainMuller, line 28: Those two are actually not used (suppose they're left over from a previous iteration)

```suggestion
```

internal/injector/builtin/yaml/go-main.yml
```diff
@@ -21,7 +21,38 @@ aspects:
...
+                  profiler.GoroutineProfile,
+                  profiler.MutexProfile,
+                ),
+                profiler.WithTags("orchestrion:true"),
+              )
+              if err != nil {
+                // TODO: is there a better reporting mechanism?
+                // The tracer and profiler already use the stdlib logger, so
+                // we're not adding anything new. But users might be using a
+                // different logger.
```
RomainMuller, line 54: That's a good question... I think it makes sense to do whatever `dd-trace-go` does... It may be interesting for it to expose a logger utility API that customers could re-configure to use their own logger instead? But I don't think this exists today... so... good enough!

RomainMuller (approved):

felixge (approved):
Thank you so much 🙇 . This LGTM.

internal/injector/builtin/yaml/go-main.yml
```diff
@@ -21,7 +21,36 @@ aspects:
...
+              // profiling.
+              err := profiler.Start(
+                profiler.WithProfileTypes(
+                  profiler.CPUProfile,
+                  profiler.HeapProfile,
+                  // Non-default profiles which are highly likely to be useful:
+                  profiler.GoroutineProfile,
+                  profiler.MutexProfile,
+                ),
+                profiler.WithTags("orchestrion:true"),
```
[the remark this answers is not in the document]
nsrip-dd: Yeah, I'll check on the status of that. It should be straightforward to add that info to profiles from the client library since the client library can tell [whether it was built with Orchestrion](https://github.com/DataDog/dd-trace-go/blob/6674e13ea04b053da120c3b83a4d79d99f2d5e5e/internal/orchestrion/orchestrion.go#L8-L16).

RomainMuller (approved):

PR #180, opened 2024-07-23
docs: fix broken pkg.go.dev links

nsrip-dd:
The links to pkg.go.dev were broken due to mismatched quotation marks,
mixing single and double quotes. This lead to the "target" and "rel"
attributes getting added to the actual URL, like

	https://pkg.go.dev/net/http#Server" target="_blank" rel="noopener

See, for example, the link to the `net/http.Server` docs on
https://datadoghq.dev/orchestrion/docs/built-in/stdlib/net-http.server/:

Screenshot 2024-07-23 at 15 04 31


RomainMuller (approved):
Good cacth!

PR #185, opened 2024-07-29
feat: support instrumentation of k8s.io/client-go

rarguelloF:
### What does this PR do?

- Add orchestrion support for `k8s.io/client-go`
- Add support for writing multiple test cases under `_integration-tests` - just need to export structs named `TestCaseXYZ`.
- Fix an issue where panics happening inside `tc.Run(t)` were being swallowed in `_integration-tests`. 

### Reviewer's Checklist

- [x] Changed code has unit tests for its functionality.


RomainMuller (approved):

_integration-tests/utils/generator/main.go
```diff
@@ -33,12 +57,51 @@ func main() {
 		return strings.Compare(lhs.Name(), rhs.Name())
 	})
 
+	dc := decorator.NewDecorator(nil)
```
RomainMuller, line 60: I'm okay with using `dave/dst` here because we already use it so much in the rest of the project; but since we don't manipulate comments much, we *could* have decided to instead use the standard `go/ast`.

rarguelloF: OK, makes sense! let me look into changing that.

PR #189, opened 2024-07-31
feat: support instrumentation of hashicorp/vault

darccio:
### What does this PR do?

Adds support for instrumenting `github.com/hashicorp/vault`.

### Reviewer's Checklist


_integration-tests/go.mod
```diff
@@ -5,7 +5,6 @@ go 1.21
 replace github.com/datadog/orchestrion => ../
 
 require (
-	github.com/datadog/orchestrion v0.0.0-00010101000000-000000000000
```
RomainMuller: It's weird this would disappear?

internal/injector/builtin/testdata/client/vault.go.snap
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/hashicorp/vault/api"
+//line <generated>:1
+	__orchestrion_vaulttrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/hashicorp/vault/api"
+)
+
+//line samples/client/vault.go:14
+func vaultClient() {
+	c, err := api.NewClient(&
+//line <generated>:1
+//line samples/client/vault.go:15
```
RomainMuller, line 20: Okay -- this is weird (but also this is not your PR's fault so don't go try to fix that).

internal/injector/builtin/yaml/api/vault.yml
```diff
@@ -0,0 +1,34 @@
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2024-present Datadog, Inc.
+%YAML 1.1
+---
```
RomainMuller, line 6: We have decided to no longer put the YAML version directive because it messes with Goland's schema validation (among other reasons). Also, now you can bind it to the schema document:

```suggestion
---
# yaml-language-server: $schema=../../../../../docs/static/schema.json
```

internal/injector/builtin/yaml/api/vault.yml
```diff
@@ -0,0 +1,34 @@
...
+meta:
+  name: github.com/hashicorp/vault/api
+  description: Official package for interacting with a Vault server.
+  icon: at-symbol
```
RomainMuller, line 10: (Optional) I think it might be a nice touch to use a shield or key icon instead for this one:

```suggestion
  icon: shield-check
```

```suggestion
  icon: key
```

_integration-tests/go.mod
darccio: No idea why. I did run `go mod tidy`. Does it happen in your local branch too?

darccio: Solved. I didn't run it again in `_integration_tests`.

_integration-tests/tests/vault/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+	"time"
+
+	"github.com/hashicorp/vault/api"
+	"github.com/stretchr/testify/require"
+	"github.com/testcontainers/testcontainers-go"
+	testvault "github.com/testcontainers/testcontainers-go/modules/vault"
+)
+
+type TestCase struct {
```
RomainMuller, line 20: This test case is not executed... it doesn't implement the interface properly as it's missing expected traces...

_integration-tests/tests/vault/main.go
```diff
@@ -0,0 +1,71 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package vault
```
rarguelloF, line 6: I think you are missing running the `go generate ./tests` so this test is actually added to the test suite.

darccio: You are right. I probably missed a `go generate ./...`.

RomainMuller (approved):

PR #193, opened 2024-07-31
chore: refactor CLI handling using `urfave/cli`

RomainMuller:
Properly model the CLI arguments using `urfave/cli`, as it supports the idiomatic Go flag format (which ,among other quirks, is ambivalent about `-flag` versus `--flag`), and allows nicely encapsulating subcommands, including opting out of any flag parsing (this is extremely useful when coding proxy commands).

This moves the `pin` command implementation to `internal/pin` from where it was before, so as to have better encapsulation.

The `ORCHESTRION_LOG_LEVEL` and `ORCHESTRION_LOG_FILE` environment variables are now also exposed as global flags on the `orchestrion` command, which provides better discoverability.


internal/cmd/toolexec.go
```diff
@@ -0,0 +1,64 @@
...
+
+		log.Tracef("Toolexec original command: %q\n", proxyCmd.Args())
+		weaver := aspect.Weaver{ImportPath: os.Getenv("TOOLEXEC_IMPORTPATH")}
+
+		if err := proxy.ProcessCommand(proxyCmd, weaver.OnCompile); err != nil {
+			return err
+		} else if err := proxy.ProcessCommand(proxyCmd, weaver.OnCompileMain); err != nil {
+			return err
+		} else if err := proxy.ProcessCommand(proxyCmd, weaver.OnLink); err != nil {
+			return err
```
eliottness, line 58: I believe we don't need the `else` here because these are returning if statements. (We probably already had this debate before)

internal/cmd/version.go
```diff
@@ -0,0 +1,37 @@
...
+			Name:    "verbose",
+			Aliases: []string{"v"},
+			Usage:   "display the version of the orchestrion binary that started this command (if different from the current)",
+			Hidden:  true,
+		},
+	},
+	Action: func(c *cli.Context) error {
+		if c.Bool("verbose") {
+			if startupVersion := ensure.StartupVersion(); startupVersion != version.Tag {
+				fmt.Printf("%s (started via %s)\n", version.Tag, startupVersion)
```
eliottness, line 30: Could we do something similar to golangci-lint:
```
❯ golangci-lint --version
golangci-lint has version 1.59.1 built with go1.22.3 from 1a55854a on 2024-06-09T18:08:33Z
```

main.go
```diff
@@ -193,16 +86,45 @@ func printUsage(cmd string) {
 	}
 }
 
-var (
-	orchestrionBinPath string // The path to the current executable file
-)
+var logLevelSet bool
```
eliottness, line 89: nit: make it a sync.Once

eliottness (commented):
Looks (almost) good to me.

eliottness (approved):

eliottness (approved):

PR #196, opened 2024-07-31
chore: upgrade all dependencies, add CODEOWNERS and SAST scanning

RomainMuller:
Further improve the OpenSSF score by upgrading all dependencies to reduce the amount of apparent vulnerabilities present there.

Additionally, add a `CODEOWNERS` file with obvious contents to satisfy and OpenSSF suggestion.

Finally, runs CodeQL analysis on the codebase.


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @DataDog/apm-go @DataDog/asm-go
```
eliottness, line 1: Wouldn't it be better to use @dd-trace-go-guild to also include people fron profiling ?

eliottness (changes requested):

.github/CODEOWNERS
RomainMuller: They have a tiny, tiny integration and I would rather avoir imposing notification fatigue on them

go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/datadog/orchestrion
 
-go 1.21
+go 1.22.0
```
eliottness, line 3: I thought you did not wanted to switch to 1.22 because of Go Support Policy

RomainMuller: 😡 this is `go mod tidy` doing things nobody ever asked it to 😡

eliottness (approved):

PR #202, opened 2024-08-02
chore: dedicated GitHub Workflow to keep go.mod/go.sum up-to-date

RomainMuller:
Instead of using Dependabot, which does inspects and udpates each `go.mod` file separately, issuing one PR for each file; use a custom GitHub Workflow to update all three `go.mod` files in a single PR.

This also includes provisions to ignore `v1.999.*` releases of the tracer library, which are experimental releases (that are likely to be retracted in the future).

This also introduces requirement to pin external actions to the relevant full-length-SHA, as per OpenSSF recommendations.


.github/workflows/deps-update.yml
```diff
@@ -0,0 +1,116 @@
...
+          go-version: '1.21' # Minimum supported go release
+          cache-dependency-path: '**/*.sum'
+
+      # Select the latest available version of gopkg.in/DataDog/dd-trace-go.v1, while ignoring all
+      # the `v1.999.*` versions, which are experimental pre-releases.
+      - name: Determine gopkg.in/DataDog/dd-trace-go.v1 version
+        id: dd-trace-go
+        run: |-
+          set -euo pipefail
+          version=$(go list -f '{{range .Versions}}{{.}}{{"\n"}}{{end}}' -m -versions gopkg.in/DataDog/dd-trace-go.v1 | grep -v -E '^v1\.999\.' | tail -n 1)
```
eliottness, line 34: Are we confident enough about the version ordering to be able to use ˋtail -n 1` ?

.github/workflows/deps-update.yml
```diff
@@ -0,0 +1,116 @@
...
+      # Passing "go@<version>" to "go get -u" ensures no dependencies get upgraded to a release that
+      # does not support that specific go release.
+      - name: "Update: Project root"
+        run: go get -t -u go@1.21 gopkg.in/DataDog/dd-trace-go.v1@${{ steps.dd-trace-go.outputs.version }} ./... && go mod tidy
+      - name: "Update: Samples"
+        run: go get -t -u go@1.21 gopkg.in/DataDog/dd-trace-go.v1@${{ steps.dd-trace-go.outputs.version }} ./... && go mod tidy
+        working-directory: samples
+      - name: "Update: Integration Tests"
+        run: go get -t -u go@1.21 gopkg.in/DataDog/dd-trace-go.v1@${{ steps.dd-trace-go.outputs.version }} ./... && go mod tidy
+        working-directory: _integration-tests
```
eliottness, line 46: Why put the `./...` thingy ?

.github/workflows/deps-update.yml
```diff
@@ -0,0 +1,116 @@
...
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+
+      - name: Set up Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5
+        with:
+          go-version: 1.21 # Minimum supported go release
+          cache-dependency-path: '**/*.sum'
```
eliottness, line 86: This does not seem to be required
```suggestion
```

.github/workflows/deps-update.yml
RomainMuller: According to `go list` documentation:

>The `-versions` flag causes list to set the Module's Versions field
to a list of all known versions of that module, ordered according
to semantic versioning, earliest to latest.

.github/workflows/deps-update.yml
RomainMuller: If you don't put `./...` and pass _any_ argument (`package@query`), then `go get -u` will only touch those things you specified and their transitive dependencies... We wanna update everything that's in use.

.github/workflows/deps-update.yml
RomainMuller: You're correct! This is from when I wanted to `go install` `ghcommit` until I found about their action ^^

eliottness (approved):

PR #208, opened 2024-08-05
feat: add log/slog instrumentation

nsrip-dd:
### What does this PR do?

Add instrumentation for `log/slog`. Implemented using dd-trace-go's
`log/slog.Handler` wrapper to wrap any handler passed to `log/slog.New`.

This commit does not instrument the default `log/slog.Logger`. We could
do that, for example, by adding some logic like this to `func main()` if
`log/slog` is imported by package main:

```go
slog.SetDefault(
    slog.New(
        slogtrace.WrapHandler(
            slog.Default().Handler(),
        ),
    ),
)
```

Will leave that as a followup for now.

### Motivation

I maintain an app that uses `log/slog` for logging. I'd like to eventually
instrument the app using Orchestrion, and having trace-log correlation is very
useful for debugging.

### Reviewer's Checklist

- [x] Changed code has unit tests for its functionality.


felixge (approved):
Awesome 🤩 . This looks good to me - but I'm not super familiar with orchestrion internals.

_integration-tests/tests/slog/main.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		if ok, _ := regexp.Match(`dd.trace_id=\d+`, line); !ok {
+			t.Errorf("no trace ID")
+		}
+	}
+}
+
+func (tc *TestCase) Teardown(t *testing.T) {}
+
+func (*TestCase) ExpectedTraces() trace.Spans { return trace.Spans{} }
```
RomainMuller, line 73: Possibly for follow-up...

Could we somehow verify instrumentation actually wrapped the handler? I get that this doesn't show up in traces... but maybe some other way?

I'm guessing the logger could write to some buffer owned by the TestCase, and the `tearDown` function could verify this has been adequately modified?

RomainMuller (approved):

RomainMuller (approved):

PR #213, opened 2024-08-07
fix: build fails when `-cover` and/or `-covermode` are passed

RomainMuller:
Overhaul the `Injector` API to stop using `decorator.Load` which internally relies on the use of `packages.Load` (similar to `go list`), which caused all dependencies to be built without instrumentation as part of the instrumentation build (i.e, wasted effort).

It turns out it is viable to instead use a combination of standard library APIs to construct just the required inputs:
1. use `go/parser.ParseFile` to parse the original source code to a `go/ast` tree
2. use `go/types` to type-check the source code
	- use a custom `go/importer` that looks up from the `-importcfg` file
	- extract the `Uses` map from produced `types.Info`
3. Create a `decorator.Decorator` using `gotypes.DecoratorResolver` built from the `Uses` map
4. From this point on, the previous mode of operation is fully supported

This actually simplifies the flow quite a bit, as fewer information is now necessary from the `proxy.CompileCommand` (we no longer need to infer a source tree root), and allows working with files that have been transformed by other tools (such as `go tool cover`) seamlessly.

---

This PR also introduces a new way to adjust line position information (generation of `//line` directives) to produce more accurate results. It does so by comparing line information of nodes between the newly restored AST & the original AST, and making sure adjusted line information matches; adding new directives as necessary.

---

This also updates the test runners' go runtimes to `go1.22.6` and `go1.21.13` releases, which include bug fixes that allow us to:

1. forego using `gotip` to use the `covdata` tool, as it no longer keeps an excessive amount of file descriptors open;
2. collect coverage of the integration tests themselves during ingteration test runs.

---

Fixes #214


internal/injector/lineinfo/annotation.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/dave/dst"
+	"github.com/dave/dst/decorator"
+)
+
+const generated = "<generated>"
+
+type (
+	// annotationVisitor is an ast.Visitor that adds `//line` directives to the visited nodes to
+	// adjust their logical source location so it matches those of the original `*ast.File` tree.
```
eliottness, line 22: I thought is was done by dave/dst automatically. Is this comment missing some context?

eliottness (commented):
This PR definitively needs to be broken down into multiple ones. I left only few of the original comments so you can focus on this

internal/injector/lineinfo/annotation.go
RomainMuller: No, `dave/dst` does none of this...

RomainMuller:
Superseded by #234, #231, #230, #229, #228, #227

PR #216, opened 2024-08-08
feat: support instrumentation of `github.com/{Shopify,IBM}/sarama`

darccio:
### What does this PR do?

Adds support for instrumenting `github.com/Shopify/sarama` and `github.com/IBM/sarama`.

Context propagation is handled in `WrapPartitionConsumer`, which is called by the wrapped consumer from `WrapConsumer`.

### Reviewer's Checklist


RomainMuller:
I confirm the flakiness. I've tried to switch to using `redpanda` instead of `confluent` test container, as this has a dev mode that may perform better... however this does not appear to have fixed the flakiness.

I wonder if maybe the contrib itself is flaky?

RomainMuller (approved):

PR #217, opened 2024-08-08
feat: add profiling of Orchestrion

nsrip-dd:
### What does this PR do?

Add hidden flags to enable profiling Orchestrion. Options for writing
CPU and heap profiler, as well as execution traces, with a configurable
path prefix. Enabling the profiler for the parent enables it for the
children. There is one of each enabled profile type per process, and
each file gets the PID in its name.

Note that CPU profiling adds a ~200ms delay at the end of the app to
stop profiling. This might matter more for Orchestrion since there are
many normally short-lived processes.

Execution tracing should be enabled with caution since traces can be
quite large. Even the other profile types can produce significant
amounts of data. The profiles will have redundant symbol data. The CPU
and heap profiles can be combined into a single profile to eliminate the
redundancy.

Example usage:

```
$ mkdir profiles
$ orchestrion --profile-prefix=${PWD}/profiles --cpu-profiling=true go build
# To combine the profiles:
$ go tool pprof -proto profiles/*.pprof > cpu.pprof
# To view:
$ go tool pprof -http=localhost:6060 cpu.pprof
```

### Motivation

Looking for hot spots & bottlenecks in Orchestrion

### Reviewer's Checklist

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@ import (
 	"runtime/pprof"
 	"runtime/trace"
 	"strconv"
+	"strings"
 
 	"github.com/datadog/orchestrion/internal/cmd"
 	"github.com/datadog/orchestrion/internal/jobserver/client"
@@ -26,13 +27,9 @@ const (
 	envVarOrchestrionLogFile  = "ORCHESTRION_LOG_FILE"
 	envVarOrchestrionLogLevel = "ORCHESTRION_LOG_LEVEL"
 	envVarToolexecImportPath  = "TOOLEXEC_IMPORTPATH"
-	envVarCPUProfiling        = "CPU_PROFILING"
-	envVarHeapProfiling       = "HEAP_PROFILING"
-	envVarExecutionTracing    = "EXECUTION_TRACING"
-	envVarProfilePrefix       = "PROFILE_PREFIX"
+	envVarProfilePath         = "ORCHESTRION_PROFILE_PATH"
+	envVarEnabledProfiles     = "ORCHESTRION_ENABLED_PROFILES"
 )
-
-var profilePrefix string
 
 type ctxKeyCPUProfiling struct{}
 type ctxKeyHeapProfiling struct{}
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -22,8 +26,18 @@ const (
 	envVarOrchestrionLogFile  = "ORCHESTRION_LOG_FILE"
 	envVarOrchestrionLogLevel = "ORCHESTRION_LOG_LEVEL"
 	envVarToolexecImportPath  = "TOOLEXEC_IMPORTPATH"
+	envVarCPUProfiling        = "CPU_PROFILING"
+	envVarHeapProfiling       = "HEAP_PROFILING"
+	envVarExecutionTracing    = "EXECUTION_TRACING"
+	envVarProfilePrefix       = "PROFILE_PREFIX"
```
RomainMuller, line 32: Environment variables that are CLI flag stand-in should be named `ORCHESTRION_<SHOUT_CASE_OF_FLAG_NAME>`, so the relationship is clear.

main.go
```diff
@@ -22,8 +26,18 @@ const (
...
 )
 
+var profilePrefix string
+
+type ctxKeyCPUProfiling struct{}
+type ctxKeyHeapProfiling struct{}
+type ctxKeyExecutionTracing struct{}
```
RomainMuller, line 39: I find it a little awkward that the `profilePrefix` is shipped into a package-local variable; while the three enablement variables are stored as `context.Context` values.

I think it should be possible to:
- Have all flags use a `Destination` that is a variable local to `main`
- Start all the requested profilers in a `Before` closure

main.go
```diff
@@ -125,3 +199,62 @@ func actionSetLogFile(_ *cli.Context, path string) error {
 
 	return nil
 }
+
+func profilePath(nameFormat string) string {
+	filename := fmt.Sprintf(nameFormat, os.Getpid())
+	if profilePrefix != "" {
+		filename = filepath.Join(profilePrefix, filename)
```
RomainMuller, line 206: It bothers me that this function implies there is some flag ordering incidence -- the actions are called as soon as the flags is seen/processed, so if the `-profile-prefix` flags is provided after a profiler selection flag, it won't be honored.

main.go
```diff
@@ -67,6 +81,44 @@ func main() {
...
+				Hidden:   true,
+			},
+			&cli.BoolFlag{
+				Category: "Profiling",
+				Name:     "execution-tracing",
+				EnvVars:  []string{envVarExecutionTracing},
+				Usage:    "Enable the execution tracer. Traces are recorded to file \"orchestrion-${pid}.trace\"",
+				Action:   actionExecutionTracing,
+				Hidden:   true,
+			},
```
RomainMuller, line 121: These feel to me like they could be a list argument (e.g: `-profile=heap -profile=cpu -profile=execution-tracing`)?

RomainMuller (changes requested):

main.go
nsrip-dd: Done

main.go
nsrip-dd: I got rid of the variable for the profilePrefix, and moved all the profile initialization into `Before` as suggested.

main.go
nsrip-dd: Good call, done.

main.go
nsrip-dd: Indeed. Moved this all to `Before` as suggested in your other comment, which as I understand it will run after all of the args are processed.

nsrip-dd (commented):
Thanks for the review @RomainMuller. I made several changes based on your comments.

RomainMuller (approved):

PR #223, opened 2024-08-15
fix: apply `//dd:orchestrion-enabled` to tracer-internal modules

nsrip-dd:
### What does this PR do?

The //dd:orchestrion-enabled directive aspect is currently not applied to
tracer-internal modules. dd-trace-go will rely on this directive to detect
whether it was built with Orchestrion. So, set "tracer-internal: true" for the
directive.

Nothing seems to be broken right now. I'm not sure the best way to test this, at
least in this PR. I'm working on an integration test for the profiler when
built with Orchestrion. Once that's up and running, it will break if the
profiler relies on the directive to add Orchestrion-related info. And
presumably once we start relying on goroutine-local storage for context, stuff
will break if the directive doesn't work in dd-trace-go?

### Motivation

I happened to notice this was broken while trying to add Orchestrion-related
information to the profiler uploads. The `internal/orchestrion.Enabled`
function in dd-trace-go was always returning false, even when built with
Orchestrion.

### Reviewer's Checklist


nsrip-dd:
@eliottness The integration tests are failling with this change. [Example](https://github.com/DataDog/orchestrion/actions/runs/10409684123/job/28829849035?pr=223#step:7:85). I think it's due to [this part in dd-trace-go](https://github.com/DataDog/dd-trace-go/blob/0462924d538179978007d967821fb241431d4423/internal/orchestrion/context_stack.go#L11C1-L22C17):

```
type contextStack map[any][]any

// getDDContextStack is a main way to access the GLS slot of runtime.g inserted by orchestrion. This function should not be
// called if the enabled variable is false.
func getDDContextStack() *contextStack {
	if gls := getDDGLS(); gls != nil {
		return gls.(*contextStack)
	}

	newStack := new(contextStack)
	setDDGLS(newStack)
	return newStack
}
```

The `new(contextStack)` call gives you a non-nil pointer to a nil map. So the methods later check that the receiver `s *contextStack` is not nil, but that just means the pointer isn't nil. There needs to be a `make` call to actually create the map. I think adding `*newStack = make(contextStack)` would work?

eliottness (approved):

RomainMuller:
This particular fix is also present in https://github.com/DataDog/orchestrion/pull/225

RomainMuller (approved):

RomainMuller:
This actually needs #232 before the CI will consistently pass; it is currently flaky due to re-use of improperly cleaned up `g` objects.

Issue #161, opened 2024-07-12
[BUG] Orchestrion does not register properly the use of `-C` option in pinning logic


eliottness:
**Version of orchestrion**

But present on the main branch last commit: [5130436](https://github.com/DataDog/orchestrion/commit/513043660f61d9d6f1c9bb7a14af44ba333ca901)

**Describe what happened:**

![image](https://github.com/user-attachments/assets/12279ff9-65cd-496c-8244-0e97b37760c3)

**Describe what you expected:**

I expected both commands to work but the one with `-C` to displace the call to all go commands was not properly intepreted by orchestrion while the second one works fine.

**Steps to reproduce the issue:**

```
git clone https://github.com/DataDog/orchestrion
go -C orchestrion build -o="../orchestrion.exe" .
./orchestrion.exe go -C orchestrion/_integration-tests mod tidy
```

**Additional environment details (Version of Go, Operating System, etc.):**

`go env`:
<details>

```
❯ go env
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/eliott/.cache/go-build'
GOENV='/home/eliott/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/eliott/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/eliott/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21.5'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD='/home/eliott/git/github.com/DataDog/orchestrion/_integration-tests/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
[...]

PR #230, opened 2024-08-19
fix(goflags): using `-cover` causes fingerprint mismatch

RomainMuller:
In order to properly convey coverage flags to child processes when doing injected package resolution, it is important the `-cover` flags only applies to packages selected for instrumentation by the user, otherwise some packages may be instrumented or not instrumented depending on whether they are directly resolved or a transitive dependency of a directly resolved package; and this leads to link-time fingerprint mismatches.

This implies we must make the default value of `-coverpkg` explicit if it is not provided by the user.

### DIFF
--- a/internal/goproxy/dashc.go
+++ b/internal/goproxy/dashc.go
@@ -1,0 +1,45 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package goproxy
+
+import (
+	"os"
+	"strings"
+
+	"github.com/datadog/orchestrion/internal/log"
+)
+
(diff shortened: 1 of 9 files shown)


internal/goproxy/dashc.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/datadog/orchestrion/internal/log"
+)
+
+// ProcessDashC gets the command line arguments passed to a "go" command (without "go" itself), and processes the "-C"
+// flag present at the beginning of the slice (if present), changing directories as requested, then returns the slice
+// without it.
+//
+// The "-C" flags is reuqired to be the very first argument provided to "go" commands.
+func ProcessDashC(args []string) ([]string, error) {
```
eliottness, line 20: You can probably use this to fix #161

eliottness (approved):

internal/goproxy/dashc.go
RomainMuller: Yah intentionally left for a future PR because I think it might be a little trickier than you think ^^

PR #232, opened 2024-08-20
fix: GLS slot not cleaned up after goroutine exits

RomainMuller:
### What does this PR do?

Specifically `nil` out the GLS slot when the goroutine exits, so that the underlying map can be GC'd and it is not incorrectly re-used by the next goroutine if that g is recycled from the gf list.

### Motivation

The GLS slot was not reset to nil when the goroutine exits, causing the old context to be re-
used whenever the g object was recycled from the free list.


_integration-tests/tests/aws.v1/aws.go
```diff
@@ -36,7 +36,7 @@ func (tc *TestCase) Setup(t *testing.T) {
 		ContainerRequest: testcontainers.ContainerRequest{
 			Image:        "amazon/dynamodb-local:latest",
 			ExposedPorts: []string{port},
-			WaitingFor:   wait.ForListeningPort(nat.Port(port)),
+			WaitingFor:   wait.ForHTTP("").WithStatusCodeMatcher(func(int) bool { return true }),
```
eliottness, line 39: what is this ?

eliottness (approved):

PR #236, opened 2024-08-22
WIP: add support for gocql

rarguelloF:


internal/injector/builtin/yaml/databases/gocql.yml
```diff
@@ -0,0 +1,25 @@
...
+      method-call: github.com/gocql/gocql.Cluster.CreateSession
+    advice:
+      - wrap-expression:
+          imports:
+            gocql: github.com/gocql/gocql
+            gocqltrace: gopkg.in/DataDog/dd-trace-go.v1/contrib/gocql/gocql
+          template: |-
+            func() (*gocql.Session, error) {
+              return gocqltrace.CreateTracedSession({{ .Receiver }})
+            }()
```
RomainMuller, line 25: There's a few things that could be problematic here...

- The source position of the original call is lost here; and maybe we need to have a way to "fix that" (effectively stating that the `gocqltrace.CreateTracedSession` is actually standing in for the original call.
- The `{{ .Receiver }}` thing here would actually be something like `{{ .AST.Fun.X }}`
	- This is on account that the matched node would be a `*dst.CallExpr` of which the `.Fun` is a `*dst.SelectorExpr` of which we need to drop the selector only.

PR #244, opened 2024-08-26
feat: add support for aws-sdk-go-v2

rarguelloF:
- Adds support for `github.com/aws/aws-sdk-go-v2`.
- It also adds a new parameter for the `struct-literal` join-point `match`, which allows to limit the match by the struct literal style: `value-only`, `pointer-only` and `any`. This was required for the `aws-sdk-go-v2` implementation.


_integration-tests/tests/aws.v2/base.go
```diff
@@ -0,0 +1,131 @@
...
+				"resource": "DynamoDB.ListTables",
+				"type":     "http",
+			},
+			Meta: map[string]any{
+				"aws.operation": "ListTables",
+				"aws.region":    "test-region-1337",
+				"aws_service":   "DynamoDB",
+				"http.method":   "POST",
+				// TODO: investigate why this tag is not being set
+				// "http.status_code": "200",
```
RomainMuller, line 107: Do or do not, there is no TODO

docs/static/schema.json
```diff
@@ -631,7 +631,7 @@
...
+            "markdownDescription": "The `struct-literal` join point matches struct literal expressions that create an instance of the named struct. In case of `match` being equal to `value-only` or `any`, the matched node will be of type `CompositeLit`, and in case of being `pointer-only`, it will be of type `UnaryExpr` (the node itself will be available in the `{{ .X }}` field). If `field` is specified, it only matches the literals that explicitly specify the named field.\n\nThe struct literal may have its address immediately taken (`&SomeType{/*...*/}`), and associated advice must be carefully designed to avoid breaking this (for example, wrapping it in an immediately-invoked function expression makes it impossible to take the value's address without first assigning it to a variable). For cases like this, the `match` field comes handy to limit the type of expressions that are matched.",
```
RomainMuller, line 634:
```suggestion
            "markdownDescription": "The `struct-literal` join point matches struct literal expressions that create an instance of the named struct. In case of `match` being equal to `value-only` or `any`, the matched node will be of type `CompositeLit`, and in case of being `pointer-only`, it will be of type `UnaryExpr` (the node itself will be available in the `{{ .X }}` field). If `field` is specified, it only matches the value explicitly associated to the named field named field.\n\nWhen using `match: any`, the struct literal may have its address immediately taken (`&SomeType{/*...*/}`), and associated advice must be carefully designed to avoid breaking this (for example, wrapping it in an immediately-invoked function expression makes it impossible to take the value's address without first assigning it to a variable).",
```

RomainMuller (changes requested):

RomainMuller (approved):

PR #247, opened 2024-08-27
chore: improve CodeCov configuration

RomainMuller:
Upload all reports to CodeCov at once at the end of the validation workflow, so that it stops producing mid-flight coverage reports that don't account for the entire set of reports.

Also, configure coverage collection during integration test execution.

Noticed the linter configuration was outdated and largely inert, so I fixed it up and applied A TON of linter fixes. The rule set is obviously subject to discussion & may also be updated in the future.


eliottness (approved):

eliottness:
Cannot merge because the changes made in the workflows make the `Complete` job fail in the merge_group CI because it does not take into account the new jobs that are skipped

PR #248, opened 2024-08-27
refactor(integration-tests): use container test agent

eliottness:
### What

This PR replaces all the code installing and managing the `ddapm-test-agent` process in favor of a testcontainer.

 ### Why

Benefits:
* Being able to run the testsuite as root more easily because python packages as root are broken
* The `latest` docker image is built with the last commit on main which contains https://github.com/DataDog/dd-apm-test-agent/pull/184 and remove the need for this kind of trick like here https://github.com/DataDog/orchestrion/pull/219/commits/b74fc7bb3ede1a0914b9a53d3d70f6c7e1fb7dd1
* No need to install python in the workflows


eliottness:
Docker is not available on windows and macos runners

PR #249, opened 2024-08-27
docs: improvements to the user guide

RomainMuller:
Added:
- Feature activation documentation (Tracer, Profiler, ASM)
- Trace customization (`//dd:span`, `//dd:ignore`)

Removed:
- Details from `README.md`, replaced with pointers to the user guide

### DIFF
--- a/docs/content/docs/custom-trace.md
+++ b/docs/content/docs/custom-trace.md
@@ -1,0 +1,151 @@
+---
+title: Trace Customization
+weight: 20
+prev: /docs/features
+next: /docs/built-in
+---
+
+Orchestrion offers serveral ways to control the traces produced by instrumented
+applications.
+
+## Prevent instrumentation of a section of code
+
+By default, `orchestrion` injects instrumentation _everywhere_ possible. This
+ensures users get the maximum possible coverage from their applications, as it
+removes the possibility of someone forgetting to instrument a particular call.
+
+There are however cases where you may want specific sections of your application
+to not be instrumented, either because they result in excessively verbose
+traces, or because those trace spans would be duplicated (for example, when
+using a custom-made `net/http` middleware stack).
+
+The `//dd:ignore` directive can be added anywhere in your application's code,
+and will disable all `orchestrion` instrumentation in the annotated syntax tree.
+
+For example:
+
+```go
+package demo
+
+import "net/http"
+
+//dd:ignore I don't want any of this to be instrumented, ever.
+func noInstrumentationThere() {
+  // Orchestrion will never add or modify any code in this function
+  // ... etc ...
+}
+
+func definitelyInstrumented() {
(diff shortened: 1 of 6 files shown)


docs/content/docs/custom-trace.md
```diff
@@ -0,0 +1,151 @@
...
+
+If the annotated function accepts a {{<godoc "context" "Context" >}} argument,
+that context will be used for trace propagation. Otherwise, if the function
+accepts a {{<godoc "net/http" "Request" "*">}} argument, the requests's context will
+be used for trace propagation.
+
+Functions that accept neither solely rely on _goroutine local storage_ for trace
+propagation. This means that traces may be split on _goroutine_ boundaries
+unless a {{<godoc "context" "Context" >}} or {{<godoc "net/http" "Request" "*">}}
+value is passed across.
```
nsrip-dd, line 132: Does this last sentence mean that if a context or request _is_ passed to child goroutines, that the trace _will_ be propagated? In other words, if `caller` in the example below didn't accept a context parameter, but passed one to `callee`, would the trace context be propagated?

```go
//dd:span
func caller() {
  wait := make(chan struct{}, 1)
  defer close(wait)

  ctx, cancel := context.WithTimeout(10*time.Second)
  defer cancel()
  // Is the span propagated here to callee?
  go callee(ctx, wait)
  <-wait
}
```

docs/content/docs/custom-trace.md
```diff
@@ -0,0 +1,151 @@
...
+  tracedFunction := func() {
+    // This function will be represented as a span named "spanName"
+  }
+  ```
+
+### Trace Context Propagation
+
+If the annotated function accepts a {{<godoc "context" "Context" >}} argument,
+that context will be used for trace propagation. Otherwise, if the function
+accepts a {{<godoc "net/http" "Request" "*">}} argument, the requests's context will
```
nsrip-dd, line 126:
```suggestion
accepts a {{<godoc "net/http" "Request" "*">}} argument, the request's context will
```

nsrip-dd (approved):
The profiler docs look good to me. I left one question about the trace context propagation section.

docs/content/docs/custom-trace.md
RomainMuller: Darn... I intentionally remained fuzzy on this because I hadn't tested it yet, and hoped nobody would notice 🥸

RomainMuller: But so in case the `ctx` is created _inside_ of the span, then no, it will not propagate context (because it does not have it).

Quite frankly, the more I look at this, the more I think @felixge was right in that we probably need to propagate the GLS context slot to new goroutines the way profiler flags are propagated... But then we suddenly need to ensure the slot's value is safe to use from multiple goroutines...

PR #254, opened 2024-08-30
chore: refactor documentation generation

RomainMuller:
Move the HTML template fragments to the `generator` package in order to improve the lisibility of the join points & advice codebases.

This allows using `html/template` templates instead of composing HTML into a buffer, then serializing it, and makes it easier to spot markup issues.

One caveat with this approach is that since this HTML gets embedded into a markdown document, care has to be taken to ensure that nothing gets indented more than 4 spaces deep if it is preceded by a blank line, as this is the syntax for Markdown indented code blocks (which cannot be disbaled).

Bonus features include:
- Now rendering shorter names when referring to a single type, by showing only the package name instead of the entire import path (still available on hover, as it's the `title` attribute of an `<abrr>` tag)
- Dropped a number of function options that were only used in unit tests


eliottness (approved):
I already reviewed most of it before you rebased it. LGTM :+1:

PR #256, opened 2024-09-02
feat: support gocql and add new method-call join-point

rarguelloF:
- Adds support for `gocql/gocql`.
- Adds new `method-call` join-point (continuation from https://github.com/DataDog/orchestrion/pull/245)


internal/injector/aspect/join/join.go
```diff
@@ -120,6 +121,29 @@ func (n *TypeName) Matches(node dst.Expr) bool {
...
+
+	switch typ := typ.(type) {
+	case *types.Pointer:
+		return false
+	case *types.Named:
+		obj := typ.Obj()
+		if obj.Pkg().Path() != n.path {
+			return false
+		}
+		return obj.Name() == n.name
```
eliottness, line 141:
```suggestion
		if obj.Pkg().Path() != n.path {
			return false
		}
		return obj.Pkg().Path() == n.path && obj.Name() == n.name
```

internal/injector/aspect/join/join.go
```diff
@@ -120,6 +121,29 @@ func (n *TypeName) Matches(node dst.Expr) bool {
...
+	default:
+		return false
+	}
```
eliottness, line 144:
```suggestion
	typ, ok := typ.(*types.Named)
	return ok && typ.Obj().Pkg().Path() == n.path && typ.Obj().Name() == n.name
	switch typ := typ.(type) {
	case *types.Pointer:
		return false
	case *types.Named:
		obj := typ.Obj()
		if obj.Pkg().Path() != n.path {
			return false
		}
		return obj.Name() == n.name
	default:
		return false
	}
```

PR #266, opened 2024-09-10
feat: add support for gcp pubsub

rarguelloF:
Adds support for instrumenting `cloud.google.com/go/pubsub`

Requires https://github.com/DataDog/dd-trace-go/pull/2852

PR #273, opened 2024-09-12
feat: add support for github.com/confluentinc/confluent-kafka-go/kafka (& its v2 counterpart)

darccio:
Adds support for instrumenting `github.com/confluentinc/confluent-kafka-go/kafka` and `github.com/confluentinc/confluent-kafka-go/v2/kafka`.

Tests for v2 library are commented because both libraries cannot coexist in the same package. An upcoming PR from @rarguelloF will separate the tests in a way that will keep them apart, so we'll be able to run both test suites. The v2 tests have been tested locally successfully.


_integration-tests/tests/confluentinc_confluent-kafka-go.v2/kafka.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+	tc.server, err = redpanda.Run(ctx,
+		"docker.redpanda.com/redpandadata/redpanda:v24.2.1",
+		redpanda.WithAutoCreateTopics(),
+		testcontainers.WithLogger(testcontainers.TestLogger(t)),
+		utils.WithTestLogConsumer(t),
+	)
+	if err != nil {
+		t.Skipf("Failed to start kafka test container: %v\n", err)
+	}
```
rarguelloF, line 52:
```suggestion
		utils.AssertTestContainersError(t, err)
```

this function already exists to check whether the error should mark the test as skipped or not

_integration-tests/tests/confluentinc_confluent-kafka-go/kafka.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+	tc.server, err = redpanda.Run(ctx,
+		"docker.redpanda.com/redpandadata/redpanda:v24.2.1",
+		redpanda.WithAutoCreateTopics(),
+		testcontainers.WithLogger(testcontainers.TestLogger(t)),
+		utils.WithTestLogConsumer(t),
+	)
+	if err != nil {
+		t.Skipf("Failed to start kafka test container: %v\n", err)
+	}
```
rarguelloF, line 52:
```suggestion
	utils.AssertTestContainersError(t, err)
```

same here

internal/injector/builtin/yaml/datastreams/confluentinc_kafka_v2.yml
```diff
@@ -0,0 +1,44 @@
...
+  - id: Wrap kafka.NewProducer
+    join-point:
+      function-call: github.com/confluentinc/confluent-kafka-go/v2/kafka.NewProducer
+    advice:
+      - wrap-expression:
+          imports:
+            kafkatrace: gopkg.in/DataDog/dd-trace-go.v1/contrib/confluentinc/confluent-kafka-go/kafka.v2
+            kafka: github.com/confluentinc/confluent-kafka-go/v2/kafka
+          template: |-
+            func(p *kafka.Producer, err error) (*kafkatrace.Producer, error) {
```
rarguelloF, line 38: I don't think this is a safe replacement to do, as we originally have a `*kafka.Producer` and we return a `*kafkatrace.Producer`

if the user has code like this, it would break:

```
var (
	producer *kafka.Producer
	err error
)

producer = // this replacement
```

Also, if there are any type assertions in the users code, these would break.

I think the only safe approach here is to instrument the library code itself, similar as I did here: https://github.com/DataDog/orchestrion/pull/266

cc @RomainMuller @eliottness please confirm if this is correct 🙇

eliottness: I can confirm. There is the same issue with gorilla/mux

rarguelloF:
closed in favor of https://github.com/DataDog/orchestrion/pull/320

PR #281, opened 2024-09-18
fix: CompileCommand may return extraneous files

RomainMuller:
In cases where an argument that is not positional ends with `.go`, the `CompileCommand.GoFiles` function might erroneously return that value as a source file to be compiled. This is for example the case when compiling the `github.com/nats-io/nats.go` package, as the `-p` argument has a value ending in `.go`.

To address this, this PR introduces a more reliable arguments parser that uses a generated parser (built using the standard `flag` package) based on the usage documentation found in the output of `go tool <compile|link>`, so that the flags are more clearly identified.

This replaces the old arguments parser, which is no longer used.


internal/toolexec/proxy/generator/main.go
```diff
@@ -0,0 +1,258 @@
...
+					log.Fatalln(err)
+				}
+				if _, isKnown := knownFlags[flag]; !isKnown {
+					log.Fatalf("Unknown captured flag: %q\n", flag)
+				}
+				capturedFlags[flag] = field.Names[0].Name
+			}
+		}
+	}
+
```
eliottness, line 141:
```suggestion
	if len(capturedFlags) == 0 {
		log.Fatalf("Expected fields with the 'ddflag' tag from the following struct: %q\n", typeName)
	}
```

internal/toolexec/proxy/generator/main.go
```diff
@@ -0,0 +1,258 @@
...
+		BlockFunc(func(g *jen.Group) {
+			g.Id("flagSet").
+				Op(":=").
+				Qual("flag", "NewFlagSet").
+				Call(
+					jen.Lit(command),
+					jen.Qual("flag", "ContinueOnError"),
+				)
+
+			reDefault := regexp.MustCompile(`^(.+?)(?:\s+\(default\s+(.+?)\))?$`)
```
eliottness, line 166:
```suggestion
			// reDefault looks for default values in descriptions
			// https://regex101.com/r/jDEwWQ/1
			reDefault := regexp.MustCompile(`^(.+?)(?:\s+\(default\s+(.+?)\))?$`)
```

internal/toolexec/proxy/generator/main.go
```diff
@@ -0,0 +1,258 @@
...
+			for _, spec := range flags {
+				fieldName, captured := capturedFlags[spec.Flag]
+
```
eliottness, line 169:
```suggestion
				delete(capturedFlags[spec.Flag])
```

internal/toolexec/proxy/generator/main.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			g.Line()
+
+			g.Id("err").Op(":=").Id("flagSet").Dot("Parse").Call(jen.Id("args"))
+			g.Return(
+				jen.Id("flagSet").Dot("Args").Call(),
+				jen.Id("err"),
+			)
+		})
+
```
eliottness, line 233:
```suggestion
		if len(capturedFlags) > 0 {
			log.Fatalf("Unknown flags used as 'ddflag' tag in struct: %q\n", capturedFlags)
		}
```

internal/toolexec/proxy/link.go
```diff
@@ -37,7 +44,10 @@ func parseLinkCommand(args []string) (Command, error) {
 		return nil, errors.New("unexpected number of command arguments")
 	}
 	flags := &linkFlagSet{}
-	parseFlags(flags, args[1:])
+	_, err := flags.parse(args[1:])
+	if err != nil {
+		return nil, err
+	}
```
eliottness, line 50:
```suggestion
	if _, err := flags.parse(args[1:]); err != nil {
		return nil, err
	}
```

eliottness (changes requested):
I have 2 concerns:
* Are we sure we can trash the unknown arguments just like that?
* I would expect at least 1 unit test of the code generated by the generator, on each go version in our matrix

eliottness:
For examples things have already changes on my computer:
image

eliottness:
Same diff after I upgraded from go 1.22.6 to 1.23.1:
image

RomainMuller:
The generation should be done only with the latest version of the compiler, as this could have more options than a previous one... I'm gonna look at having some meta somewhere that allows to make sure we don't "roll back" to an older compiler version.

The defaults being different based on the machine hardware/limits is a little annoying... I wonder what should be done (maybe special-case the known ones that are machine-dependent and just ignore them?)

internal/toolexec/proxy/generator/main.go
RomainMuller: This is already checked for when the capture tags are processed.

eliottness (approved):
(let's fix the CI but) LGTM

PR #291, opened 2024-09-23
chore: run ddapm-test-agent in docker

rarguelloF:
### What

This PR changes the setup of `ddapm-test-agent` from running in a python subprocess to a shared docker container. 

### Motivation

I've observed flakiness in integration tests with the following error:

```    
gen_test.go:26: 
        	Error Trace:	/home/runner/work/orchestrion/orchestrion/_integration-tests/utils/suite.go:70
        	            				/home/runner/work/orchestrion/orchestrion/_integration-tests/tests/gocql/gen_test.go:26
        	Error:      	Received unexpected error:
        	            	Get "http://127.0.0.1:32779/test/session/start?test_session_token=8dd45585-440c-446f-886e-2ec732e7756b": context deadline exceeded
        	Test:       	TestIntegration_gocql_StructLiteralPtr
```

Running the agent using testcontainers should provide a better way to manage the lifecycle of the agent process, so we don't attempt to run the test when the agent is not ready or not running at all.

This change also simplifies the setup as python is no longer required and allows to run tests concurrently using the same agent by configuring the http client used by the tracer to set the `X-Datadog-Test-Session-Token` header for every request.

I opened an alternative [PR](https://github.com/DataDog/orchestrion/pull/282) with an very basic implementation of the mock agent in Go that should provide similar benefits, in case we are not interested in the extra features the ddapm-test-agent provides.


RomainMuller (commented):
This will beak testing of Windows & macOS variants in GitHub actions (Windows only supports Windows guests, and macOS flat out does not support nested virtualization).

From this perspective, I'd rather go with #282 (it's pure go, fewer dependencies, supports all platforms we wanna target).

PR #303, opened 2024-09-30
chore: reduce allocations in injector

RomainMuller:
Use a `sync.Pool` to reduce allocations performed by the injector, and change the `functionOption` implementations so that they no longer receive the context as a pointer (which caused a heap escape); instead by-copy (the struct is 64 bytes, which is small enough).

This removes a non-trivial amount of allocations, which may have non-negligible performance impact.

Bonus -- Changed some doc linkage to use the common godoc convention, for improved IDE experience :)

---

Jira: APPSEC-54998


internal/injector/aspect/context/node.go
```diff
@@ -19,16 +21,32 @@ type NodeChain struct {
...
 		panic("cursor does not point to a child of this node")
 	}
-	return &NodeChain{
-		parent: n,
-		node:   cursor.Node(),
-		name:   cursor.Name(),
-		index:  cursor.Index(),
-	}
+
+	c, _ := pool.Get().(*NodeChain)
```
eliottness, line 35:
```suggestion
	c := pool.Get().(*NodeChain)
```
It's faster without the type check right ?

eliottness (commented):
I have 2 suggestions:
- I know it does not look trivial to do but if you feel like it, can you add a test to check if the sync.Pool is empty at the end? Like for memory leaks
- I guess we can also run the tool betteralign to check for cases where the struct layout is suboptimal

RomainMuller:
>I guess we can also run the tool betteralign to check for cases where the struct layout is suboptimal
I have done that already, but our structs are relatively small and those feel like rounding errors, basically.

>I know it does not look trivial to do but if you feel like it, can you add a test to check if the sync.Pool is empty at the end? Like for memory leaks
If used **incorrectly**, the `sync.Pool` will be empty at the end. If used **correctly**, it will contain whatever amount of values it deems reasonable.

eliottness (approved):

PR #311, opened 2024-10-01
docs: getting started with orchestrion built-in profiling

RomainMuller:
Provides a basic guide on how to get started with the built-in `--profile-path` and `--profile` flags of orchestrion when performing performance investigation.


nsrip-dd (approved):
LGTM, thanks!

PR #316, opened 2024-10-03
fix: `//dd:span` on `func main()` does not work

RomainMuller:
Since the tracer was started at the start of the `main` function, a `//dd:span` annotation there may not have been honored, depending on the order in which aspects were woven into the file.

This PR mvoes the tracer initialization to a `func init()` function, so that it happens strictly before the `main` function starts executing, making it possible to `//dd:span` the `main` function reliably.


samples/client/main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"time"
 )
 
+//dd:span
```
eliottness, line 18: If I am not mistaken the samples tests do not test if traces are outputted. Wouldn't it be better to put this in the dd:span integration test ?

eliottness (approved):
Thoughts for later: would it be better to divide this aspect into 2: one for starting the tracer and one for the profiler ? Especially since we can add multiple init functions in the same file

samples/client/main.go
RomainMuller: We cannot have a `func main()` in tests... we could *maybe* do something by doing an `orchestrion go run` in there...

eliottness: No no no. This is already an weird edge case, your explanation is enough

internal/injector/builtin/yaml/go-main.yml
```diff
@@ -18,39 +18,55 @@ aspects:
...
+                  profiler.WithTags("orchestrion:true"),
+                )
+                if err != nil {
+                  // TODO: is there a better reporting mechanism?
+                  // The tracer and profiler already use the stdlib logger, so
+                  // we're not adding anything new. But users might be using a
+                  // different logger.
+                  log.Printf("failed to start profiling: %s", err)
+                }
+                defer profiler.Stop()
```
nsrip-dd, line 69: This will stop the profiler as soon as `func init()` returns, meaning we won't get profiling. We could either keep the profiler start in `main`, or remove the call to `profiler.Stop`. I think either would work fine.

I still plan to set up some end-to-end testing of Orchestrion + profiling which would catch this, just had other priorities recently.

nsrip-dd (changes requested):

internal/injector/builtin/yaml/go-main.yml
RomainMuller: Oh yikes, obviously! Good thing you're noticing this...

nsrip-dd (approved):

PR #320, opened 2024-10-07
feat: support confluent-kafka-go v1 and v2

rarguelloF:
Requires: https://github.com/DataDog/dd-trace-go/pull/2907


internal/toolexec/proxy/compile.go
```diff
@@ -60,6 +60,17 @@ func (cmd *CompileCommand) AddFiles(files []string) {
...
 }
 
+// SetLang sets the -lang flag used for compiling. The given string should be a valid
+// go lang version (e.g. "go1.23").
+func (cmd *CompileCommand) SetLang(lang string) {
+	cmd.Flags.Lang = lang
+	for i, arg := range cmd.args {
+		if strings.HasPrefix(arg, "-lang") {
+			cmd.args[i] = "-lang=" + lang
+		}
```
RomainMuller, line 70: I think this needs to be made so that it works with all arguments passing formats...

```suggestion
		if strings.HasPrefix(arg, "-lang=") {
			cmd.args[i] = "-lang=" + lang
			break
		} else if arg == "-lang" && i+1 < len(cmd.args) {
			cmd.args[i+1] = lang
			break
		}
```

RomainMuller (approved):
Ship only once it references a tagged release of dd-trace-go.

PR #325, opened 2024-10-08
chore: dependency upgrade workflow fails

RomainMuller:
The procedure to create the PR requires the branch is checked out, but the work tree is dirty at time of doing so; as the commit is created from the dirty files via the GitHub API. Clean up the working copy before switching to the newly created commit.


RomainMuller:
Confirmed it works - it created #326

PR #340, opened 2024-10-16
feat(injector: stdlib): instrument testing.[M, T, B] with CI Visibility

tonyredondo:
Adds support for instrumenting the Go `testing` package (`testing.M`, `testing.T` and `testing.B`) with CI Visibility.

The integration is disabled by default while we are testing it, in order to enabled it we need to set `DD_CIVISIBILITY_ENABLED=true`


internal/injector/builtin/yaml/go-main.yml
```diff
@@ -19,15 +19,27 @@ aspects:
...
           template: |-
+            //go:linkname __dd_civisibility_isCiVisibilityEnabled gopkg.in/DataDog/dd-trace-go.v1/internal/civisibility/integrations/gotesting.isCiVisibilityEnabled
+            func __dd_civisibility_isCiVisibilityEnabled() bool
+            
             func init() {
-              tracer.Start(tracer.WithOrchestrion(map[string]string{"version": {{printf "%q" Version}}}))
+              // Only initialize if is not a test process (testdeps.ImportPath is set only on test processes since go1.9 in the first init function)
+              // and if CI Visibility has not been disabled (by the kill switch).
+              // For a test process the ci visibility instrumentation will initialize the tracer
+              isATestProcess := len(testdeps.ImportPath) != 0
```
RomainMuller, line 39: Im' not extremely fond of this (but if there is no other ways...).

What catches my attention is this section of the `go list` documentation:
> The -test flag causes list to report not only the named packages
but also their test binaries (for packages with tests), to convey to
source code analysis tools exactly how test binaries are constructed.
The reported import path for a test binary is the import path of
the package followed by a ".test" suffix, as in "math/rand.test".
When building a test, it is sometimes necessary to rebuild certain
dependencies specially for that test (most commonly the tested
package itself). The reported import path of a package recompiled
for a particular test binary is followed by a space and the name of
the test binary in brackets, as in "math/rand [math/rand.test]"
or "regexp [sort.test]". The ForTest field is also set to the name
of the package being tested ("math/rand" or "sort" in the previous
examples).

Hence in that cas I wonder if the toolchain even signals that `main` function as being in a `main` package in that case?

internal/injector/builtin/yaml/go-main.yml
```diff
@@ -19,15 +19,27 @@ aspects:
...
+            - gopkg.in/DataDog/dd-trace-go.v1/internal/civisibility/integrations/gotesting
           imports:
             tracer: gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
+            testdeps: testing/internal/testdeps
           # Note: it is valid to have multiple func init() in a single compile unit (e.g, `.go` file), in which case
           # they get executed in declaration order. This means it's okay for us to add a new init function if there is
           # already one in the file, but as it currently is appended AFTER all other declarations in the file, it means
           # that it will be executed last (tracing contents of previous init functions will not be possible).
           template: |-
+            //go:linkname __dd_civisibility_isCiVisibilityEnabled gopkg.in/DataDog/dd-trace-go.v1/internal/civisibility/integrations/gotesting.isCiVisibilityEnabled
```
RomainMuller, line 32: I'm curious why all these functions are internal & un-exported from the tracer library. They feel like they'd be useful or even necessary for someone to implement CI viz for a custom testing harness?

internal/injector/builtin/yaml/civisibility/testing.yml
```diff
@@ -0,0 +1,212 @@
...
+      all-of:
+        - import-path: testing
+        - function-body:
+            function:
+              - name: SkipNow
+              - receiver: '*testing.common'
+    advice:
+      - prepend-statements:
+          template: |-
+            __dd_civisibility_instrumentSkipNow({{ .Function.Receiver }})
```
RomainMuller, line 212: Please add a newline at end of file

internal/injector/builtin/yaml/go-main.yml
tonyredondo: We will do that but in the future, we want to add a manual api to the library so customer could write their own "instrumentation" on their testing framework of preference, but we are not yet there, one of the pushback is thinking everything ahead to avoid breaking changes on dd-trace-go.

So the plan is to have a few instrumentations to test this internal api first.

internal/injector/builtin/yaml/civisibility/testing.yml
```diff
@@ -0,0 +1,212 @@
+# Unless explicitly stated otherwise all files in this repository are licensed
+# under the Apache License Version 2.0.
+# This product includes software developed at Datadog (https://www.datadoghq.com/).
+# Copyright 2023-present Datadog, Inc.
+---
+# yaml-language-server: $schema=../../../../../docs/static/schema.json
+meta:
+  name: go-testing
```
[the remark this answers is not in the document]
tonyredondo: Currently in all other tracers we have CI Visibility stuff kind of isolated, for a couple of reasons: First it changes the mode of the tracer behaviour, other writers, other type of payloads, different transports, etc, so is not a simple new integration for the tracer it self. The second because we modify a lot of the behaviour we tend to group all this to a single folder and be friendly with the codeowners file. I did the same because this will be the first integration under the CI Visibility umbrella but I don't think it will the only one.

internal/injector/builtin/yaml/civisibility/testing.yml
tonyredondo: done in [95dc6c3](https://github.com/DataDog/orchestrion/pull/340/commits/95dc6c315c0a7f7a4249e0fea2c57b8fd55c52ff)

internal/injector/builtin/yaml/go-main.yml
tonyredondo: I did change the approach to use the new [`testing.Testing()` api ](https://github.com/golang/go/commit/7f38067acb738c43d870400dd648662d31456f5f#diff-b3e6126779b5ec9d3d6cea7cc54054ba78f4d7a0a6248d9e458bbd9b3d72fce3R659)added in Go 1.21 on [bedccb0](https://github.com/DataDog/orchestrion/pull/340/commits/bedccb03e7e5586f4e3256415c113fbe160a01f5)

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/tinylib/msgp/msgp"
+)
+
+var ciVisibilityPayloads mockPayloads
+
+func TestMain(m *testing.M) {
```
RomainMuller, line 25: Optional, but it might be a good idea to validate that Orchestrion did indeed instrument the package (for otherwise, you'll get a ton of assertion failures that are actually just "there is no instrumentation here").

There's a magic comment you can use to have a package-level `const` (or `var` but I'd recommend the `const` form here) boolean variable turn to `true` when build w/ orchestrion.

See: https://datadoghq.dev/orchestrion/docs/built-in/directive/orchestrion-enabled/

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+			gzipReader, err := gzip.NewReader(r.Body)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			defer gzipReader.Close()
+
+			// Convert the message pack to json
+			jsonBuf := new(bytes.Buffer)
+			_, err = msgp.CopyToJSON(jsonBuf, gzipReader)
```
RomainMuller, line 130:
```suggestion
			var jsonBuf bytes.Buffer
			_, err = msgp.CopyToJSON(&jsonBuf, gzipReader)
```

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+func enableCiVisibilityEndpointMock() *httptest.Server {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/api/v2/citestcycle" {
+			fmt.Printf("mockapi: test cycle payload received.\n")
+
+			// first we need to read the body
+			// then we need to unzip the body
+			// then we need to convert the body to json
+			// then we need to decode the json
```
RomainMuller, line 119: I'd have those comments spread out with the code that effectively implements them. Also, I'd prefer them to be a little bit more specific:

```suggestion
			// first we need to read the body
			// then we need to gunzip the body
			// then we need to convert the body from msgpack to json
			// then we need to parse the json
```

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+			testWithSubtestsChild2).
+		HasCount(0)
+
+	// All previous checks will cause panic if they fail so we can safely exit with 0 here
+	os.Exit(0)
+}
+
+func enableCiVisibilityEndpointMock() *httptest.Server {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/api/v2/citestcycle" {
```
RomainMuller, line 113: This condition should be flipped, so the "happy case" is un-indented...

Really what this function is:

```suggestion
		if r.URL.Path != "/api/v2/citestcycle" {
			w.WriteHeader(http.StatusBadRequest)
			return
		}
```

I also wonder if that should be an HTTP 404 here; but it's not really important...

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+			fmt.Printf("mockapi: test cycle payload received.\n")
+
+			// first we need to read the body
+			// then we need to unzip the body
+			// then we need to convert the body to json
+			// then we need to decode the json
+
+			gzipReader, err := gzip.NewReader(r.Body)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
RomainMuller, line 123: It would probably be useful to log that error... 

```suggestion
			if err != nil {
				log.Println(err)
				w.WriteHeader(http.StatusInternalServerError)
```

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+				return
+			}
+			defer gzipReader.Close()
+
+			// Convert the message pack to json
+			jsonBuf := new(bytes.Buffer)
+			_, err = msgp.CopyToJSON(jsonBuf, gzipReader)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
RomainMuller, line 132:
```suggestion
			if err != nil {
				log.Println(err)
				w.WriteHeader(http.StatusInternalServerError)
```

_integration-tests/tests/civisibility/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+				return
+			}
+
+			var payload mockPayload
+			err = json.Unmarshal(jsonBuf.Bytes(), &payload)
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
RomainMuller, line 139:
```suggestion
			if err != nil {
				log.Println(err)
				w.WriteHeader(http.StatusInternalServerError)
```

internal/injector/builtin/yaml/civisibility/testing.yml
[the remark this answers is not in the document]
RomainMuller: That makes sense to me... But I'd still just name this `testing` (Orchestrion only touches go, so it's a bit of a stutter)

```suggestion
  name: testing
```

internal/injector/builtin/yaml/go-main.yml
RomainMuller, line 1: For posterity -- as discussed separately, I hope we can actually make it so that the `main.main` aspect is not woven into generated `*_testmain.go` files somehow, and this would make these changes unnecessary, I think.

That'll be investigated separately though; as this is more of a "purity" kind of thing, and functionally the end-result should be the same.

internal/injector/builtin/yaml/civisibility/testing.yml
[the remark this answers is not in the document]
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: oh that's nice, done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

_integration-tests/tests/civisibility/main_test.go
tonyredondo: done in [92ee768](https://github.com/DataDog/orchestrion/pull/340/commits/92ee768360c07a61c2311485193b2236c93be0b1)

internal/injector/builtin/yaml/go-main.yml
tonyredondo: yes skipping the `*_testmain.go` from the main integration will be a better approach in the future.

RomainMuller (approved):

PR #345, opened 2024-10-21
release: v0.9.3

RomainMuller:
Introduces the initial version of CI visibility instrumentation.


eliottness:
Should we add something to the README ?

eliottness (approved):

RomainMuller:
> Should we add something to the README or the docs ?

I'm going to suggest this to Tony, but for now I just wanted to unblock dogfooding so we can "real-life-use-case" vet it before advertising it too loudly

PR #351, opened 2024-10-22
fix(injector: redigo) fix bad argument type from []redis.DialOption to []any

eliottness:
Error:
```
❯ orchestrion go test -tags dynamic -args integration -v ./... -count 1
# go.ddbuild.io/dd-source/x/libs/go/redis
/Users/tony.redondo/go/pkg/mod/go.ddbuild.io/dd-source/x@v0.0.0-20241021135812-b13784d37c9c/libs/go/redis/connector.go:38: cannot use dialOptions (variable of type []"github.com/gomodule/redigo/redis".DialOption) as []interface{} value in argument to __orchestrion_redigo.Dial
exit status 2
FAIL    github.com/DataDog/dd-go/ci-app/apps/ci-test-commit-summarizer/testsummarizer [build failed]
```

Fix `github.com/gomodule/redigo/redis` aspects where adding options ahead of making calls would fail the build because of the differences of signatures between the original and replaced function:

Original DialURL function:
```go
func DialURL(rawurl string, options ...redis.DialOption) (redis.Conn, error)
```

dd-trace-go DialURL function:
```go
func DialURL(rawurl string, options ...any) (redis.Conn, error)
```

Therefore we need to cast the args before sending them to dd-trace-go


darccio (approved):

internal/injector/builtin/yaml/databases/redigo.yml
```diff
@@ -16,23 +16,71 @@ aspects:
...
+          imports:
+            redigo: github.com/gomodule/redigo/redis
+            redigotrace: gopkg.in/DataDog/dd-trace-go.v1/contrib/gomodule/redigo
+          template: |-
+            func() (redis.Conn, error) {
+              {{ $len := len .AST.Args }}
+              {{ if le $len 2 }}
+                return redigotrace.Dial({{ index .AST.Args 0 }}, {{ index .AST.Args 1 }})
+              {{ else }}
+                opts := {{ index .AST.Args 2 }}
```
RomainMuller, line 29: We should unit test these aspect with all possible call styles, because I suspect this isn't gonna work properly in some... This is a variadic function, so it can be:

- `redigo.Dial(x)` (OK)
- `redigo.Dial(x, y)` (Suspect not ok)
- `redigo.Dial(x, y, z, a, b, c)` (Suspect not ok)
- `redigo.Dial(x, arr...)` (OK)

eliottness: There was indeed an issue. I fixed it @RomainMuller

RomainMuller (approved):

PR #352, opened 2024-10-23
feat: support github.com/twitchtv/twirp

rarguelloF:
Adds support for `github.com/twitchtv/twirp`


internal/injector/builtin/yaml/rpc/twirp.yml
```diff
@@ -0,0 +1,37 @@
...
+            {{- .AST.Type -}}{
+              {{- $hasField := false -}}
+              {{ range .AST.Elts }}
+              {{- if eq .Key.Name "Hooks" }}
+              {{- $hasField = true -}}
+              {{- end -}}
+              {{ . }},
+              {{ end -}}
+              {{- if not $hasField -}}
+              Hooks: twirptrace.NewServerHooks(),
```
eliottness, line 35: There is the `ChainHooks` function that should be able to help in case the field is already set

eliottness (changes requested):
Would you mind doing a `samples/server/twirp.go` example file ?

internal/injector/builtin/yaml/rpc/twirp.yml
rarguelloF: nice! didn't know about it

eliottness (approved):

PR #357, opened 2024-10-24
feat: code templates can declare required go langage level

RomainMuller:
In certain cases, it is not practical to write aspects using go syntax that is known to work with all go versions. While this is in general not an issue (since `orchestrion` only supports the latest 2 go releases), it can still cause issues when injecting code into modules with a `go.mod` file that declares an antiquated `go` version (e.g: `go 1.13`).

In such cases, go 1.22 and newer toolchains will instruct the compiler to process the source code in a way that ensures compatibility with that antiquated language level, prohibiting use of newer language features.

To protect ourselves from this issue, the `template` objects can now be configured to demand a certain minimum language level to be available, which will cause compile units into which that template generated code to have the compiler's `-lang` flag set to that level (unless it was already set at a higher level).

This might have consequences on the behavior of the compiled code (e.g, the change of `for range` variable bindings introduced in go 1.22); but we currently do not deal with this particular issue, as we assume it will seldom have any practical impact in our use-cases.


internal/injector/aspect/context/golang.go
```diff
@@ -0,0 +1,95 @@
...
+
+func ParseGoLang(lang string) (GoLang, error) {
+	// Validate the language level value...
+	if !strings.HasPrefix(lang, "go1.") {
+		return GoLang{}, fmt.Errorf(`invalid go language level (expected e.g, "go1.18"): %q`, lang)
+	}
+	for _, r := range lang[4:] {
+		if !unicode.IsDigit(r) {
+			return GoLang{}, fmt.Errorf(`invalid go language level (expected e.g, "go1.18"): %q`, lang)
+		}
```
rarguelloF, line 30: the package `go/version` has some utilities to work with go version strings.

So probably you can replace this code with just `version.IsValid(lang)`

internal/injector/aspect/context/golang.go
```diff
@@ -0,0 +1,95 @@
...
+	if left == right {
+		return 0
+	}
+
+	if left.IsAny() {
+		// Blank is lower than any other version...
+		return -1
+	}
+
+	return semver.Compare(left.semver, right.semver)
```
rarguelloF, line 76: you can use `go/version.Compare()` that is already prepared to work with go version strings (just need to do `version.Lang(rawString)` before using it).

rarguelloF (commented):
Overall LGTM, just sorking with the go version strings can be simplified using `go/version` package instead of `golang.org/x/mod/semver`.

docs/content/contributing/aspects/code-templates.md
```diff
@@ -44,15 +44,29 @@ imports:
   tracer: gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer
 ```
 
-### Links
```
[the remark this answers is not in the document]
RomainMuller: It was never there to begin with! It's only supported for `inject-declarations`...

internal/injector/aspect/context/golang.go
RomainMuller: Ohhhhhhhh :)

internal/injector/aspect/advice/call.go
```diff
@@ -42,6 +42,7 @@ func (a *appendArgs) Apply(ctx context.AdviceContext) (bool, error) {
 		if err != nil {
 			return false, err
 		}
+		ctx.EnsureMinGoLang(t.Lang)
```
eliottness, line 45: From what I see all 4 advice implementation where you added `ctx.EnsureMinGoLang(t.Lang)` have the CodeCov warning onto them. Is this expected?

internal/injector/aspect/context/golang.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"fmt"
+	"go/version"
+
+	"gopkg.in/yaml.v3"
+)
+
+// GoLang represents a go language level. It's a string of the form "go1.18".
+type GoLang struct {
```
eliottness, line 16: Would you mind adding a "version" in there? Like `GoVersion` or `GoLanguageVersion` ? The first read of this PR was verry confusing

internal/injector/aspect/context/golang_test.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+func TestSetSetAtLeast(t *testing.T) {
```
eliottness, line 17:
```suggestion
func TestSetAtLeast(t *testing.T) {
```

internal/injector/aspect/context/golang.go
RomainMuller: I agree; but since it's internal anyway, it's okay for it to be there for now... Would move it later if needed.

internal/injector/aspect/advice/call.go
RomainMuller: That is not expected (and not the case anymore AFAICT).

eliottness (approved):

PR #370, opened 2024-10-30
feat: support github.com/sirupsen/logrus

rarguelloF:


internal/injector/builtin/yaml/logs/logrus.yml
```diff
@@ -0,0 +1,89 @@
...
+        match: value-only
+    advice:
+      - wrap-expression:
+          imports:
+            logrus: github.com/sirupsen/logrus
+          template: |-
+            func(logger logrus.Logger) logrus.Logger {
+              logger.AddHook(&logrus.DDContextLogHook{})
+              return logger
+            }({{ . }})
```
RomainMuller, line 89: I don't love how this appears to be making more copies than strictly necessary; but I reckon I might be overly pessimistic with how the compiler will treat that, and I don't think this is a line that ought to be on a hot path anyway...

RomainMuller (approved):

PR #372, opened 2024-10-30
fix(ensure): packages.Load().Module is nil

### LINKED ISSUE #371: [BUG] version check SEGV in toolexec mode on go 1.23
**Version of orchestrion**

Latest & main

**Describe what happened:**

The call `tools/packages.Load` will return with no errors with `pkg.Module` being `nil`. This being not checked create this error:

```
❯ go test -v -tags=integration,buildtag -shuffle=on -toolexec "$PWD/../orchestrion toolexec" ./tests/redigo
# github.com/docker/docker/errdefs
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xdeaed6]

goroutine 1 [running]:
github.com/DataDog/orchestrion/internal/ensure.goModVersion({0x0?, 0xc00023bc38?})
        /home/eliott/dd/orchestrion/internal/ensure/requiredversion.go:150 +0x116
github.com/DataDog/orchestrion/internal/ensure.requiredVersion(0x10fb8bf?, 0x11548e0, 0x1155108, {0xc0001b8000, 0xe, 0xe})
        /home/eliott/dd/orchestrion/internal/ensure/requiredversion.go:66 +0x6a
github.com/DataDog/orchestrion/internal/ensure.RequiredVersion(...)
        /home/eliott/dd/orchestrion/internal/ensure/requiredversion.go:44
github.com/DataDog/orchestrion/internal/pin.init.0()
        /home/eliott/dd/orchestrion/internal/pin/pin.go:146 +0x68
```

I guess we expect orchestrion execution to not happen in this case because we have no way to make sure we are running the right version of orchestrion.

This issue only happen on certain conditions:
* Need at least go 1.23
* Need to use orchestrion in TOOLEXEC mode (meaning with `-toolexec`, not with `orchestrion go ...`)
* Only some packages of the `github.com/docker/docker` repository are reproductiing this issue

Then running any integration test trying to build `github.com/docker/docker` reproduce the issue

eliottness:
Fixes #371 

See https://github.com/golang/go/issues/65816

### DIFF
--- a/internal/ensure/requiredversion.go
+++ b/internal/ensure/requiredversion.go
@@ -147,6 +147,7 @@ func goModVersion(dir string) (moduleVersion string, moduleDir string, err error
 		return "", "", errors.Join(errs...)
 	}
 
+	// Shouldn't happen but does: see https://github.com/golang/go/issues/65816
 	if pkg.Module == nil {
 		return "", "", fmt.Errorf("no module information found for package %q", pkg.PkgPath)
 	}
--- a/internal/ensure/requiredversion.go
+++ b/internal/ensure/requiredversion.go
@@ -147,6 +147,10 @@ func goModVersion(dir string) (moduleVersion string, moduleDir string, err error
 		return "", "", errors.Join(errs...)
 	}
 
+	if pkg.Module == nil {
+		return "", "", fmt.Errorf("no module information found for package %q", pkg.PkgPath)
+	}
+
 	if pkg.Module.Replace != nil {
 		// If there's a replace directive, that's what we need to be honoring instead.
 		return pkg.Module.Replace.Version, pkg.Module.Replace.Dir, nil
--- a/internal/ensure/requiredversion.go
+++ b/internal/ensure/requiredversion.go
@@ -147,7 +147,9 @@ func goModVersion(dir string) (moduleVersion string, moduleDir string, err error
 		return "", "", errors.Join(errs...)
 	}
 
-	// Shouldn't happen but does: see https://github.com/golang/go/issues/65816
+	// Shouldn't happen but does when the current working directory is not
+	// part of a go module's source tree.
+	// See: https://github.com/golang/go/issues/65816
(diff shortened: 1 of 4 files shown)


internal/ensure/requiredversion.go
```diff
@@ -147,6 +147,11 @@ func goModVersion(dir string) (moduleVersion string, moduleDir string, err error
 		return "", "", errors.Join(errs...)
 	}
 
+	// Shouldn't happen but does: see https://github.com/golang/go/issues/65816
```
RomainMuller, line 150:
```suggestion
	// Shouldn't happen but does when the current working directory is not
	// part of a go module's source tree.
	// See: https://github.com/golang/go/issues/65816
```

internal/ensure/requiredversion.go
```diff
@@ -147,6 +147,11 @@ func goModVersion(dir string) (moduleVersion string, moduleDir string, err error
...
+	if pkg.Module == nil {
+		return "", "", fmt.Errorf("no module information found for package %q", pkg.PkgPath)
```
RomainMuller, line 152: Interesting this shows as un-tested...

eliottness: This must be linked to the fact that we don't have this issue in CI but I do on my machine. I can't really understand why because I mapped exactly the go version on my machine with the CI :shrug:

RomainMuller: I think the difference is on your local machine you probably did `go install github.com/DataDog/orchestrion` so it can be found under `GOPATH`, but it's not true on CI... (so `packages.Load` actually returns an error).

RomainMuller (approved):

internal/ensure/requiredversion.go
eliottness: Wholly resetting the GOPATH does not change a thing :/

PR #378, opened 2024-11-05
attempt to fix flaky tests

rarguelloF:
- Extract and reuse redis container initialization into `utils.StartRedisTestContainer`
- Move code in testcases that needed the container and could fail from `Setup` -> `Run` - the reason is if this code fails, we usually stopped the test with `require.NoError`, and doing this inside `Setup` makes the `Teardown` to not run and leave the container running.
- Improved wait strategies for both redis and kafka, which seems to be the only tests that are flaky.


_integration-tests/tests/redigo/redigo.go
```diff
@@ -95,7 +63,10 @@ func (tc *TestCase) Run(t *testing.T) {
 
 	client, err := tc.Pool.GetContext(ctx)
 	require.NoError(t, err)
-	defer func() { require.NoError(t, client.Close()) }()
+	defer func() { assert.NoError(t, client.Close()) }()
```
RomainMuller, line 66: I'm surprised the linter doesn't complain about using `assert` to verify the absence of an error... But cool :)

_integration-tests/utils/testcontainers.go
```diff
@@ -75,6 +88,39 @@ func StartKafkaTestContainer(t *testing.T) (*kafka.KafkaContainer, string) {
...
+			),
+		),
+	)
+	AssertTestContainersError(t, err)
+
+	connStr, err := container.ConnectionString(ctx)
+	require.NoError(t, err)
+
+	redisURL, err := url.Parse(connStr)
+	require.NoError(t, err)
```
RomainMuller, line 119: So I guess if these fail; the test will just bail out right away and not clean up the container...

RomainMuller (approved):
As discussed, we should follow-up with the use of `t.Cleanup`.

PR #382, opened 2024-11-06
chore: use t.Cleanup() for integration tests cleanup

rarguelloF:
Use `t.Cleanup` to make sure the cleanup logic is registered right after the resource is created and always gets executed when the test finishes.

### DIFF
--- a/_integration-tests/tests/pgx/pgx.go
+++ b/_integration-tests/tests/pgx/pgx.go
@@ -45,12 +45,18 @@ func (tc *TestCase) Setup(t *testing.T) {
 		),
 	)
 	utils.AssertTestContainersError(t, err)
+	utils.RegisterContainerCleanup(t, tc.container)
 
 	dbURL, err := tc.container.ConnectionString(ctx, "sslmode=disable")
 	require.NoError(t, err)
 
 	tc.conn, err = pgx.Connect(ctx, dbURL)
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		defer cancel()
+		assert.NoError(t, tc.conn.Close(ctx))
+	})
 }
 
 func (tc *TestCase) Run(t *testing.T) {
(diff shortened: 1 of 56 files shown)


_integration-tests/tests/pgx/pgx.go
```diff
@@ -45,12 +45,18 @@ func (tc *TestCase) Setup(t *testing.T) {
...
 
 	dbURL, err := tc.container.ConnectionString(ctx, "sslmode=disable")
 	require.NoError(t, err)
 
 	tc.conn, err = pgx.Connect(ctx, dbURL)
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		defer cancel()
+		assert.NoError(t, tc.conn.Close(ctx))
```
eliottness, line 58: Do we use the require or assert package ? Can we normalize that ?

_integration-tests/tests/sql/sql.go
```diff
@@ -25,6 +26,9 @@ func (tc *TestCase) Setup(t *testing.T) {
 	var err error
 	tc.DB, err = sql.Open("sqlite3", "file::memory:")
 	require.NoError(t, err)
+	t.Cleanup(func() {
+		assert.NoError(t, tc.DB.Close())
```
eliottness, line 30: Why replace require by the assert package ?

eliottness (approved):
Skimmed through everything. LGTM 👍

_integration-tests/tests/pgx/pgx.go
rarguelloF: we can use both:
- `require` is an assertion followed by `t.FailNow`: I typically use this when this check is required for the test to continue (e.g. the rest of the test cannot even run if the container fails to start, or if I'm gonna be looking for specific indexes in a slice, I would use `require.Len` first on the slice so I don't see panics in the output, which make the output harder to read).
- `assert` is a normal assertion, the test will continue and fail, and you will see in the output all of the ones that failed.

In this case this is the last thing in the test and its a cleanup, it doesn't make sense that we `panic` / stop everything else in the test just because the cleanup failed, marking the test as failed is enough.

_integration-tests/tests/sql/sql.go
rarguelloF: I think we don;t need to use `require` in the cleanup function - as I mentioned in the previous comment, the test (if there is anything remaining) can continue despite of this failing.

_integration-tests/tests/pgx/pgx.go
eliottness: Did not know that. Thanks for explaining

PR #384, opened 2024-11-07
fix(goenv.GOMOD): never rely on the environment variable

RomainMuller:
The value of `go env GOMOD` changes depending on the working directory
where the command is executed. Instead of relying on the environment
variable, always spawn `go env GOMOD` in an explicitly selected
directory.


internal/ensure/requiredversion.go
```diff
@@ -128,10 +129,16 @@ func requiredVersion(
...
+	if err != nil {
+		return "", "", err
+	}
+
 	cfg := &packages.Config{
-		Dir:  dir,
-		Mode: packages.NeedModule,
-		Logf: func(format string, args ...any) { log.Tracef(format+"\n", args...) },
+		BuildFlags: []string{"-modfile", gomod},
+		Dir:        dir,
```
eliottness, line 139: Does this fix something ? We have no reason to believe that passing the `(*packages.Config).Dir` and specifying the arg  `-modfile` changes anything no ?

internal/pin/pin.go
```diff
@@ -91,7 +91,7 @@ func defaultOrchestrionToolGo() *dst.File {
 // PinOrchestrion applies or update the orchestrion pin file in the current
 // working directory, according to the supplied [Options].
 func PinOrchestrion(opts Options) error {
-	goMod, err := goenv.GOMOD()
+	goMod, err := goenv.GOMOD(".")
```
eliottness, line 94:
```suggestion
	goMod, err := goenv.GOMOD("")
```

Previous behaviour was to use `""` and I am not knowledgeable enough to know if `"." ` is 100%  exactly the same thing on platforms like windows

internal/ensure/requiredversion.go
RomainMuller: It allows us to have a more graceful error if the working directory isn't a go module; and passing `-modfile` allows us to explicitly use the `gomod` value instead of just dropping it (we'd only be interested in the error).

eliottness (approved):

PR #387, opened 2024-11-08
feat: new `built.WithOrchestrion` constant

RomainMuller:
Expose a `built.WithOrchestrion` constant so that usages where it is possible to take a direct dependency on `github.com/DataDog/orchestrion` is possible are simpler; as a readily available variable is provided.

This reduces the amount of places where the `//dd:orchestrion-enabled` directive is necessary, which may improve compilation speeds (as directive join points may be dispandious to evaluate).


internal/injector/aspect/join/declaration.go
```diff
@@ -49,7 +97,26 @@ func (i *valueDeclaration) AsCode() jen.Code {
 	return jen.Qual(pkgPath, "ValueDeclaration").Call(i.TypeName.AsCode())
 }
 
+var symbolNamePattern = regexp2.MustCompile(`\A(.+)\.([\p{L}_][\p{L}_\p{Nd}]*)\z`, regexp2.ECMAScript)
```
eliottness, line 100: Where was this taken from? If you created it yourself, can you add a regex101 link ?

Also: why start using `regex2` ?

internal/injector/aspect/join/expression.go
```diff
@@ -59,17 +54,26 @@ func (i *functionCall) Matches(ctx context.AspectContext) bool {
 }
 
 func (i *functionCall) AsCode() jen.Code {
-	return jen.Qual(pkgPath, "FunctionCall").Call(jen.Lit(i.ImportPath + "." + i.Name))
+	return jen.Qual(pkgPath, "FunctionCall").Call(jen.Lit(i.ImportPath), jen.Lit(i.Name))
 }
 
-var funcNamePattern = regexp.MustCompile(`\A(?:(.+)\.)?([^.]+)\z`)
+var funcNamePattern = regexp2.MustCompile(`\A(?:(.+)\.)?([\p{L}_][\p{L}_\p{Nd}]*)\z`, regexp2.ECMAScript)
```
eliottness, line 60: Same here about the regex :)

internal/injector/builtin/yaml/directive/orchestrion-enabled.yml
```diff
@@ -17,8 +17,10 @@ aspects:
     tracer-internal: true # This is safe to apply in the tracer itself.
     join-point:
       all-of:
-        - directive: dd:orchestrion-enabled
         - value-declaration: bool
+        - one-of:
+            - declaration-of: github.com/DataDog/orchestrion/built.WithOrchestrion
+            - directive: dd:orchestrion-enabled
```
eliottness, line 23: Can you change the name of the aspect ? because it's not self-encompassing anymore

eliottness (changes requested):
How about putting the `built` package inside another package named like `api` or `runtime` or `lib` so we can regroup packages that can be called by orchestrion deps ?

internal/injector/aspect/join/declaration.go
RomainMuller: Go's `regex` does not support unicode categories >_>

RomainMuller: My bad, it actually supports those we need... :)

runtime/doc.go
```diff
@@ -0,0 +1,10 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+// Package runtime provides the run-time API of orchestrion packages. It
+// includes features such as some run-time healpers used by injected
```
eliottness, line 7:
```suggestion
// includes features such as some run-time helpers used by injected
```

internal/toolexec/aspect/oncompile.go
```diff
@@ -31,9 +31,13 @@ const (
...
-		regexp.MustCompile(`^github\.com/[Dd]ata[Dd]og/orchestrion(?:/.+)?$`): neverWeave,
-		regexp.MustCompile(`^gopkg\.in/DataDog/dd-trace-go.v1(?:/.+)?$`):      weaveTracerInternal,
-		regexp.MustCompile(`^github\.com/DataDog/go-tuf/client$`):             neverWeave,
+		// Carefully allow injection under the `github.com/DataDog/orchestrion/runtime` package, but
+		// nowhere else in the github.com/DataDog/orchestrion modules. The regex could be smaller if the
+		// [regexp.Regexp] supported negative lookahead... But there we are emulating it instead.
+		// See: https://regex101.com/r/kexx3q/1 (NB)
+		regexp.MustCompile(`^github\.com/[Dd]ata[Dd]og/orchestrion(?:/(?:[^r].*|r[^u].*|ru[^n].*|run[^t].*|runt[^i].*|runti[^m].*|runtim[^e].*|runtime[^/].*))?$`): neverWeave,
+		regexp.MustCompile(`^gopkg\.in/DataDog/dd-trace-go.v1(?:/.+)?$`):                                                                                           weaveTracerInternal,
+		regexp.MustCompile(`^github\.com/DataDog/go-tuf/client$`):                                                                                                  neverWeave,
```
eliottness, line 40: I believe we can simply use `strings.HasPrefix` for now right ? This regex looks like its making you sad

eliottness (approved):
Looks (almost) Good To Me :+1:

internal/toolexec/aspect/oncompile.go
RomainMuller: I had a problem, I used regexp to try and solve the problem. I now have a `/problem/`...

PR #391, opened 2024-11-12
feat: injector config loader API

RomainMuller:
Adds a new `internal/injector/config` package that provides features to load injector configuration, and also to inspect whether packages contain such configuration (so it can be used to implement `orchestrion pin`'s `-prune` and `-validate` features.


internal/injector/config/config-yml.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, fmt.Errorf("open %q: %w", filename, err)
+	}
+	defer file.Close()
+
+	// In validation mode, we will pre-parse the YAML into a [yaml.Node] tree,
+	// which can then be cheaply decoded into a value type that validation
+	// supports; and then re-decoded into the actual data structure.
+	// This dance is significantly cheaper (both in time & allocations) than doing
+	// a full blown [yaml.Decoder.Decode] twice (as it internally transits through
+	// the [yaml.Node] representation anyway).
```
eliottness, line 115: I don't believe this "dance" is necessary because not only are we doing it only once per build, the validate mode will only be called on `orchestrion validate || pin` right ? How about simply dividing parsing and validation ?

internal/injector/config/config-go.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return &configGo{imports: imports, yaml: ymlCfg, pkgPath: pkg.PkgPath}, nil
+}
+
+// loadGoFile loads configuration from the specified go file. Returns nil if the
+// file had already been loaded previously, or if the configuration is empty.
+func (l *Loader) loadGoFile(filename string) ([]Config, error) {
+	if !l.markLoaded(filename) {
+		return nil, nil
+	}
```
eliottness, line 63: I believe this could cause a issue because `loadGoPackage` does not handle the case where `loadGoFile` returns `nil, nil`. I believe putting the `markLoaded` on packages paths instead of filenames would make us able to make it go higher in the call stack and make things easier to manage.

internal/injector/config/config-yml.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+const OrchestrionYML = "orchestrion.yml"
+
+// loadYMLFile loads configuration from the specified directory.
+func (l *Loader) loadYMLFile(dir string, name string) (*configYML, error) {
+	filename := name
+	if !filepath.IsAbs(name) {
+		filename = filepath.Join(dir, name)
+	}
```
eliottness, line 26: Why do this on yml files and not go files ? Is there a test case covering this ?

internal/injector/config/config-yml.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+
+			cfg, err := l.loadGoPackage(pkgs[0])
+			if err != nil {
+				return nil, maskErrNotExist(err)
+			}
+			if cfg == nil {
+				// Empty, nothing to do...
+				continue
+			}
```
eliottness, line 61: This case can't happen I believe but the case where everything inside `cfg` is empty is the one that is supposed to be covered. I will let you choose whenever you want to change caller or callee side

internal/jobserver/buildid/version.go
```diff
@@ -114,7 +114,7 @@ func collectModules(pkg *packages.Package, modules map[string]*moduleInfo, known
 			modules[pkg.Module.Path] = info
 		}
 
-		for _, files := range [...][]string{pkg.GoFiles, pkg.EmbedFiles, pkg.OtherFiles} {
+		for _, files := range [...][]string{pkg.GoFiles, pkg.EmbedFiles} {
```
eliottness, line 117: why ?

internal/injector/config/config-go.go
```diff
@@ -0,0 +1,141 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/DataDog/orchestrion/internal/injector/aspect"
+	"golang.org/x/tools/go/packages"
+)
+
+const OrchestrionToolGo = "orchestrion.tool.go"
```
eliottness, line 22:
```suggestion
const OrchestrionToolGoFilename = "orchestrion.tool.go"
```
If it was only a local const I would not mind but you look like you replaced all off occurences with the variable and the name is not-so clear I think.

internal/injector/config/schema.json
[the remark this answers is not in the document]
RomainMuller: This is written by hand (at least at the moment).

instrument/orchestrion.yml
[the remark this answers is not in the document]
RomainMuller: That package contains code that is being used by some of the instrumentation (it could be moved with the aspects though).

internal/injector/builtin/generated_deps.go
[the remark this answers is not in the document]
RomainMuller: That should not be necessary... We'll need a way to validate the required dependencies are present, but shouldn't need to automatically make them happen.

internal/jobserver/buildid/version.go
RomainMuller: Oh oops, I was over-zealous!

internal/injector/config/config-yml.go
RomainMuller: `loadGoFile` is _always_ called with an absolute filename... Whereas `loadYMLFile` is (at least currently) always called with a relative filename (and it's called from the Go loader, so it double-checks for safety).

Those lines are covered by tests.

internal/injector/config/config-go.go
RomainMuller: If `loadGoFile` returns `nil, nil`, the slice of "imported packages" is `nil` which is the same as an empty slice... So that does not need to be handled specifically.

Doing `markLoaded` on package paths would work if the `orchestrion.yml` files were only able to refer to go packages, but they're able to refer to specific YAML documents with an arbitrary filename, and we still want to load each such YAML document exactly once.

internal/injector/config/config-yml.go
RomainMuller: "Deep" validation implies parsing, so it's a little painful/annoying to separate these.

eliottness (approved):

PR #392, opened 2024-11-13
feat: mechanism to fingerprint arbitrary data

RomainMuller:
Provides a facility to efficiently fingerprint the injector
configuration.


internal/fingerprint/fingerprint.go
```diff
@@ -0,0 +1,89 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package fingerprint
+
+import (
+	"crypto/sha512"
```
eliottness, line 9: This means we will have to build with boring crypto no ?

internal/injector/aspect/advice/code/template.go
```diff
@@ -221,6 +221,14 @@ func (t *Template) AsCode() jen.Code {
 	)
 }
 
+func (t *Template) Hash(h *fingerprint.Hasher) error {
+	return h.Named(
+		"template",
+		fingerprint.Map(t.Imports, func(k string, v string) (string, fingerprint.String) { return k, fingerprint.String(v) }),
+		fingerprint.String(t.Source),
+	)
+}
```
eliottness: It's missing the `t.GolangVersion` right ?

eliottness (approved):
Added 2 commits instead of doing more comments. Feel free to review them or to merge directly

internal/fingerprint/fingerprint.go
eliottness: Well anyway this is not production code so it should be ok anyway

PR #393, opened 2024-11-14
feat: dynamically load injector configuration

RomainMuller:
Use the `config.Loader` to dynamically load the injector configuration
everywhere, and not just in `orchestrion pin`.


internal/toolexec/aspect/oncompile.go
```diff
@@ -79,7 +84,7 @@ func (w Weaver) OnCompile(cmd *proxy.CompileCommand) error {
 
 		case weaveTracerInternal:
 			log.Debugf("Enabling tracer-internal mode for %q\n", w.ImportPath)
-			shortList := make([]aspect.Aspect, 0, len(aspects))
+			shortList := make([]*aspect.Aspect, 0, len(aspects))
 			for _, aspect := range aspects {
 				if aspect.TracerInternal {
 					shortList = append(shortList, aspect)
```
eliottness, line 90: We should be able to simply use `slices.DeleteFunc` here now no ?

samples/go.mod
```diff
@@ -91,7 +91,6 @@ require (
 	github.com/aws/aws-sdk-go-v2/internal/v4a v1.3.23 // indirect
 	github.com/aws/aws-sdk-go-v2/service/cloudfront v1.41.0 // indirect
 	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.37.0 // indirect
-	github.com/aws/aws-sdk-go-v2/service/ec2 v1.176.0 // indirect
```
eliottness, line 94: Why ?

internal/injector/builtin/generator/doc.tmpl
```diff
@@ -19,10 +19,10 @@
         {{ `{{<iconSVG "search-circle">}}` }} Join Point
       </span>
       <div class="root hextra-card-subtitle hx-font-normal hx-px-4 hx-mb-4 hx-mt-2">
-        {{- "\n" }}{{ render .Aspect.JoinPoint | trim }}
+        {{- "\n" }}{{ render .JoinPoint | trim }}
       </div>
     </div>
-    {{- range .Aspect.Advice }}
+    {{- range .Advice }}
```
eliottness, line 25: Was this broken ? Where was the templace execution changes in this PR ?

internal/toolexec/aspect/oncompile.go
RomainMuller: I guess, but maybe later?

samples/go.mod
RomainMuller: No idea... It just is no longer used... ¯\_(シ)_/¯

internal/injector/builtin/generator/doc.tmpl
RomainMuller: The underlying struct is no longer nested because the `ID` field was moved up to the `aspect.Aspect` struct.

eliottness (approved):

PR #394, opened 2024-11-14
feat: new `test-main` join point

RomainMuller:
The `test-main` join point allows to only/never inject into nodes
contained in the `main` package generated by `go test`. This is useful
to avoid starting the tracer & profiler during tests, which simplifies
the CI visibility aspects a little bit.


tonyredondo (approved):
👏🏻 Thanks!

internal/injector/builtin/yaml/stdlib/runtime.yml
```diff
@@ -46,3 +46,30 @@ aspects:
     advice:
       - prepend-statements:
           template: getg().__dd_gls = nil
+
+  # Temporarily hack tracer.SpanFromContext so that it no longer returns true if
+  # the Span found from GLS is actully a noop span.
```
rarguelloF, line 51:
```suggestion
  # Temporarily hack tracer.SpanFromContext so that it no longer returns true if
  # the Span found from GLS is actually a noop span.
  # This might happen when a library creates spans using the OTel SDK.  
```

[nit] just adding a little bit more context for the future

eliottness (approved):

Issue #400, opened 2024-11-18
[BUG] cannot use __orchestrion_instrument.WrapHandlerFunc


tonyredondo:
**Version of orchestrion**

0.9.4

**Describe what happened:**

Received this build error:

```
config/service/http_health_forwarder.go:157: cannot use __orchestrion_instrument.WrapHandlerFunc(func(w http.ResponseWriter, r *http.Request) {…}) (value of type "net/http".HandlerFunc) as handlerFunc value in return statement
```

function:
```go

func healthCheckBuilder(serviceName string, healthClient *httpHealthClient) handlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		// ...
	}
}

```

**Describe what you expected:**

A normal build execution

**Steps to reproduce the issue:**

**Additional environment details (Version of Go, Operating System, etc.):**

rarguelloF:
@tonyredondo Thanks for reporting the issue!

Could you please share the `handlerFunc` type? To make sure I accurately reproduce the issue.

tonyredondo:
> [@tonyredondo](https://github.com/tonyredondo) Thanks for reporting the issue!
> 
> Could you please share the `handlerFunc` type? To make sure I accurately reproduce the issue.

shared offline.

PR #403, opened 2024-11-19
fix: net/http server-side instrumentation

rarguelloF:
Resolves https://github.com/DataDog/orchestrion/issues/173, resolves https://github.com/DataDog/orchestrion/issues/400

This PR changes `net/http` server-side instrumentation aspects to trace the library itself (similar approach as we are doing today with the client-side instrumentation). The hook where we inject is the `http.Server.Serve` method, which should be the common method that is always called when the http server is started (and at this point, the Handler should  already be set).

This fixes problems in the existing instrumentation where we:
- Don't trace custom `http.Handler` implementations (https://github.com/DataDog/orchestrion/issues/173).
- We double-trace `http.HandlerFunc(func(w http.ResponseWriter, r *http.Request))` declarations.
- We might even break the build (as described in https://github.com/DataDog/orchestrion/issues/400).
- Patterns using assignment like `srv.Handler = myHandler` are not traced.

As a side-effect, this approach removes the ability to use `orchestrion:ignore` comments to skip tracing on http servers (which happens for other aspects as well where we instrument the library code).


nsrip-dd (approved):
Thanks! The general approach makes sense to me. I haven't reviewed this PR deeply. I'll leave that to the Orchestrion experts. I tried this PR out locally on a Go HTTP server we use in the profiling backend that uses ServeMux (the one that motivated #173). I confirmed that the handler gets instrumented :)

internal/toolexec/aspect/oncompile.go
```diff
@@ -62,10 +62,56 @@ var weavingSpecialCase = []specialCase{
 	{path: "github.com/DataDog/go-tuf/client", prefix: false, behavior: neverWeave},
 }
 
-func (w Weaver) OnCompile(cmd *proxy.CompileCommand) error {
+func (w Weaver) OnCompile(cmd *proxy.CompileCommand) (result error) {
```
RomainMuller, line 65: ⚠️ I'm going to do a refactor of this monstrosity tomorrow before we can merge this.

RomainMuller (changes requested):

RomainMuller (approved):

PR #419, opened 2024-11-25
fix(grpc): use itnerceptor chaining API

### LINKED ISSUE #416: [BUG]  Grpc interceptor strategy does not work if there is an existing interceptor
In Golang, you can only call grpc.UnaryInterceptor once, given that the orchestrion adds it's own interceptor, if the user had their own interceptor the code will fail with 

```
panic: The unary server interceptor was already set and may not be reset.

goroutine 1 [running]:
main.main.UnaryInterceptor.func14(0x10c0e7f01?)
        /Users/cvolantesilva/go/pkg/mod/google.golang.org/grpc@v1.68.0/server.go:439 +0x4f
google.golang.org/grpc.(*funcServerOption).apply(0x8?, 0x8?)
        /Users/cvolantesilva/go/pkg/mod/google.golang.org/grpc@v1.68.0/server.go:212 +0x19
google.golang.org/grpc.NewServer({0xc000799e98, 0x3, 0x10c8dbde2?})
        /Users/cvolantesilva/go/pkg/mod/google.golang.org/grpc@v1.68.0/server.go:661 +0x117
```

**Version of orchestrion**
0.9.4

**Describe what happened:**
Application fails to start

Application starts

Add custom interceptors in the code, ex:

`gs = grpc.NewServer(
			grpc.UnaryInterceptor(...))`

**Additional environment details (Version of Go, Operating System, etc.):**

RomainMuller:
There can be only one interceptor set by
`grpc.[With](Unary|Stream)Interceptor`, and a runtime error occurs if an interceptor is already defined and a second is attempted using the same API. Switch to using the chaining API, which appends to the existing list, and co-operates cleanly with the non-chained interceptor API.

Also, adjust the gRPC test to verify adequate behavior occurs.

Fixes #416


_integration-tests/tests/grpc/grpc.go
```diff
@@ -31,17 +32,50 @@ func (tc *TestCase) Setup(t *testing.T) {
...
+			return handler(ctx, req)
+		}),
+	)
 	helloworld.RegisterGreeterServer(tc.Server, &server{})
 
 	go func() { assert.NoError(t, tc.Server.Serve(lis)) }()
 	t.Cleanup(func() {
 		tc.Server.GracefulStop()
+		assert.True(t, interceptedDirect.Load(), "original interceptor was not called")
+		assert.True(t, interceptedChain.Load(), "original chained interceptor was not called")
```
darccio, line 56: @RomainMuller Are the messages correct? If `intercepted*` are `true`, doesn't it mean they were called? The same for the same messages below.

RomainMuller: The message is for when it fails... But I'm a litle ambivalent as to whether they should mention what should've happened or what was observed that is a failure

darccio: Oh, you are right! I usually go with something like `original interceptor expected a call, it didn't happen`. Anyway, it's fine as long as it's understandable in the testing logs' context.

darccio (approved):

PR #422, opened 2024-11-26
chore: add go.mod in `_docs` and `_tools`

RomainMuller:
Make sure the `_docs` and `_tools` directories are not part of the download package for the main orchestrion module by adding a go.mod file in them.


.github/workflows/validate.yml
```diff
@@ -246,7 +244,7 @@ jobs:
           path: ${{ runner.temp }}/benchmarks.txt
       - name: Format Report
         run: |-
-          go run golang.org/x/perf/cmd/benchstat  \
+          go _tools run golang.org/x/perf/cmd/benchstat  \
```
eliottness, line 247:
```suggestion
          go -C _tools run golang.org/x/perf/cmd/benchstat  \
```

eliottness (approved):

PR #431, opened 2024-11-28
chore: resume integration testing Windows

RomainMuller:
Added specific build tag exclusions to avoid building tests on Windows and macOS in GitHub CI that are known to not work there (due to not being able to use docker for testcontainers).


_integration-tests/utils/generator/main.go
```diff
@@ -6,31 +6,44 @@
...
 	flag.Usage = func() {
 		_, _ = fmt.Fprintf(os.Stderr, "Usage: %s <dir>\n", os.Args[0])
 		flag.PrintDefaults()
 	}
+
+	tmpl, err := template.New("").Parse(templateText)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	fileTemplate = tmpl
```
eliottness, line 46:
```suggestion
var (
	//go:embed "testfile.tmpl"
	templateText string
	fileTemplate = template.Must(template.New("").Parse(templateText))
)

func init() {
	flag.Usage = func() {
		_, _ = fmt.Fprintf(os.Stderr, "Usage: %s <dir>\n", os.Args[0])
		flag.PrintDefaults()
	}
```

_integration-tests/utils/generator/main.go
```diff
@@ -57,83 +70,124 @@ func main() {
...
-				Params(jen.Id("t").Op("*").Qual("testing", "T")).
-				Block(
-					// utils.RunTest(t, new(TestCase))
-					jen.Qual(utilsPkg, "RunTest").Call(jen.Id("t"), jen.New(jen.Id(tc))),
-				).
-				Line().Empty()
+		testDir := path.Join(root, pkg.Name())
+		testData := parseCode(testDir)
+		if len(testData.Cases) == 0 {
+			continue
```
eliottness, line 77: Why not return an error there ? There is definitively an error if a test package has no tests
```suggestion
			log.Fatalf("Test package %q has no tests cases", pkg.Name())
```

eliottness (approved):

_integration-tests/utils/generator/main.go
RomainMuller: There is one directory that contains no test (I guess I should move that around instead of putting it under the carpet)

PR #435, opened 2024-11-28
feat(injector): early filtering of aspects

eliottness:
### What

This PR adds two new functions to the `join.Point` interface:
```go
// MatchType is an enumeration of the possible outcomes of a join point
type MatchType byte

const (
	// Unknown indicates that the join point cannot determine whether it matches the given context
	Unknown MatchType = '?'
	// Match indicates that the join point may match the given context
	Match MatchType = 'Y'
	// NeverMatch indicates that the join point DOES NOT match the given context
	NeverMatch MatchType = 'N'
)

// PackageMayMatch determines whether the join point may match the given package data from the importcfg file.
PackageMayMatch(ctx *context.PackageMayMatchContext) MatchType

// FileMayMatch determines whether the join point may match the given file raw content
FileMayMatch(ctx *context.FileMayMatchContext) MatchType
```

Which shoudl help reduce drastically the number of packages we parse/decorate and the number of aspects we run each time


internal/injector/parse.go
```diff
@@ -44,17 +49,28 @@ func parseFile(fset *token.FileSet, filename string) (*ast.File, error) {
...
 
-	astFile, err := parser.ParseFile(fset, mappedFilename, file, parser.ParseComments)
+	fileContent, err := io.ReadAll(file)
+	if err != nil {
+		return nil, nil, fmt.Errorf("reading %q: %w", filename, err)
+	}
+
+	aspectsPerFile := make([]*aspect.Aspect, len(aspects))
+	for i, a := range aspects {
+		aspectsPerFile[i] = a
```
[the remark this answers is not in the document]
eliottness: Yes but actually no ⛔

internal/injector/aspect_filter.go
```diff
@@ -0,0 +1,53 @@
...
+// - For each file, we start reading it and storing its content in a []byte.
+// - If we hit no limits, we check if the content contains the filter.
+// - If all the strings from the (*Aspect).ImpliesContent() are present in the file, we keep the aspect for this file
+// - If not, we remove the aspect from the list of aspects to run on this file
+// - After all aspects are processed for a file, transform the []byte back to an io.ReadCloser and store it in the result map
+// - If any limit was hit, we stop the filtering on this file and return it as is in the result map
+func contentContainsFilter(aspects []*aspect.Aspect, fileContent []byte) []*aspect.Aspect {
+	aspectsCopy := make([]*aspect.Aspect, len(aspects))
+	for i, a := range aspects {
+		aspectsCopy[i] = a
```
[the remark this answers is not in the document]
eliottness: Yes but actually no ⛔

internal/injector/builtin/integration_test.go
```diff
@@ -69,21 +70,33 @@ func Test(t *testing.T) {
...
+				ImportMap:  importMap,
 			}
 
 			files, err := filepath.Glob(filepath.Join(pkgDir, "*.go"))
 			require.NoError(t, err)
 
-			results, _, err := inj.InjectFiles(files)
+			copyAspects := make([]*aspect.Aspect, len(aspects))
+			for i, a := range aspects {
+				copyAspects[i] = a
```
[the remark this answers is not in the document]
eliottness: Yes but actually no ⛔

internal/injector/injector.go
```diff
@@ -93,26 +102,27 @@ func (i *Injector) InjectFiles(files []string) (map[string]InjectedFile, context
...
 		resultGoLang context.GoLangVersion
 		resultMu     sync.Mutex
 	)
 
-	wg.Add(len(astFiles))
-	for idx, astFile := range astFiles {
-		go func(idx int, astFile *ast.File) {
+	wg.Add(len(parsedFiles))
+	for _, parsedFile := range parsedFiles {
+		//fmt.Printf("filename: %s nb aspects: %d\n", filename, len(aspectsPerFile[filename]))
```
eliottness, line 112:
```suggestion
```

internal/injector/check.go
```diff
@@ -29,7 +32,18 @@ func (i *Injector) typeCheck(fset *token.FileSet, files []*ast.File) (types.Info
...
+	for i, file := range files {
+		astFiles[i] = file.AstFile
+	}
+
+	if err := checker.Files(astFiles); err != nil {
+		// This is a workaround for the fact that the Go type checker does not return a specific unexported error type
+		// TODO: Ask better error typing from the Go team for the go/types package
+		if strings.Contains("package requires newer Go version", err.Error()) {
+			return types.Info{}, fmt.Errorf("orchestrion was built with Go version %s but package %q requires a newer go version, please reinstall and pin orchestrion with a newer Go version: type-checking files: %w", runtime.Version(), i.ImportPath, err)
+		}
```
RomainMuller, line 45: Should have a test exercising this code path... otherwise it's just good intentions.

internal/injector/imports.go
```diff
@@ -7,13 +7,28 @@ package injector
...
+// packageFilterAspects filters out aspects that imply imports not present in the import map.
+func (i *Injector) packageFilterAspects(aspects []*aspect.Aspect) []*aspect.Aspect {
+	ctx := &may.PackageContext{
+		ImportPath: i.ImportPath,
+		ImportMap:  i.ImportMap,
+		TestMain:   i.TestMain,
+	}
+	return slices.DeleteFunc(aspects, func(a *aspect.Aspect) bool {
+		return a.JoinPoint.PackageMayMatch(ctx) == may.CantMatch
+	})
```
RomainMuller, line 29: The implementation is modifying the `aspects` slice in-place, so the documentation should mention this.

Alternatively -- the copy that is done upstream could be done in there instead; which would remove this concern.

internal/injector/injector.go
```diff
@@ -64,6 +65,13 @@ type (
 		Filename string
 	}
 
+	parameters struct {
```
RomainMuller, line 68: That name is a little opaque... I'd prefer something a little more specific?

internal/injector/injector.go
```diff
@@ -74,18 +82,26 @@ type (
...
-	astFiles, err := parseFiles(fset, files)
+	parser := parse.NewParser(fset, len(files))
+	parsedFiles, err := parser.ParseFiles(files, aspects)
 	if err != nil {
 		return nil, context.GoLangVersion{}, err
 	}
 
-	typeInfo, err := i.typeCheck(fset, astFiles)
+	if len(parsedFiles) == 0 {
+		log.Debugf("no files to inject in %s after filtering on imports and files\n", i.ImportPath)
```
RomainMuller, line 100:
```suggestion
		log.Debugf("No files to inject in %q after filtering on imports and files\n", i.ImportPath)
```

internal/injector/injector_test.go
```diff
@@ -35,6 +35,8 @@ type testConfig struct {
 	SyntheticReferences map[string]typed.ReferenceKind `yaml:"syntheticReferences"`
 	GoLang              context.GoLangVersion          `yaml:"required-lang"`
 	Code                string                         `yaml:"code"`
+	ImportPath          string                         `yaml:"import-path"`
+	ImportMap           map[string]string              `yaml:"import-map"`
```
RomainMuller, line 39: This can probably just be a `[]string`... This is just a set, and modeling as a `map[string]string` is heavier than necessary

```suggestion
	ImportMap           []string              `yaml:"import-map"`
```

internal/injector/parse/mapped_filename.go
RomainMuller, line 1: I'd prefer we avoid naming files with an `_<suffix>.go` tail unless `<suffix>` describes a build constraint...

Either call this `mappedfilename.go` or `mapped-filename.go`

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	"os"
+	"slices"
+	"sync"
+	"sync/atomic"
+
+	"github.com/DataDog/orchestrion/internal/injector/aspect"
+	"github.com/DataDog/orchestrion/internal/injector/aspect/may"
+)
+
+const maxBytesEagerness = 1 << 19 // 512 KiB
```
RomainMuller, line 24: Please document what this implies / is used for.

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	name       string
+	mappedName string
+	content    []byte
+}
+
+type File struct {
+	Name    string
+	AstFile *ast.File
+	Aspects []*aspect.Aspect
+}
```
RomainMuller, line 36: Documentation comments, please :)

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	fset *token.FileSet // thread-safe data structure
+
+	// rawFiles is an intermediary data structure to store the raw content of the files before parsing them.
+	rawFiles []rawFile
+
+	// filesBytesCount is the sum of the bytes of all files parsed so far.
+	filesBytesCount atomic.Uint64
+
+	// eagerness is a flag that is set to true if at least one file has been parsed and requires all files to be parsed.
+	eagerness atomic.Bool
```
RomainMuller, line 48: Would name `mustParseAll` instead?

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// parsedFiles is what is returned by ParseFiles.
+	parsedFiles []File
+
+	wg sync.WaitGroup
+
+	errs   []error
+	errsMu sync.Mutex
```
RomainMuller, line 56: Could use an [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) here?

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func NewParser(fset *token.FileSet, nbFiles int) *Parser {
+	p := &Parser{
+		fset:        fset,
+		rawFiles:    make([]rawFile, nbFiles),
+		parsedFiles: make([]File, nbFiles),
+	}
+
+	p.eagerness.Store(true)
```
RomainMuller, line 66: I'm confused because this appears to contradict the documentation of `eagerness` above (it's `true`, and yet no file has been parsed yet?)

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return p
+}
+
+// ParseFiles return either zero files or all files parsed with their respective aspects
+func (p *Parser) ParseFiles(files []string, aspects []*aspect.Aspect) ([]File, error) {
```
RomainMuller, line 72: That documentation is a little lazy - it should explain in which cases it returns none versus all the parsed files.

I'm also not a huge fan of the modality of the API here, but I also see the benefit there (saves from re-traversing the set of files to see if anything has aspects)... Maybe it should return all files, always, but also return a boolean indicating whether any file has any aspect?

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// If we arrived here, this means we need to parse all files anyway because the type-checking pass will need them.
+	p.parseMissingFiles()
+	p.wg.Wait()
+
+	return p.parsedFiles, errors.Join(p.errs...)
+}
+
+// aspectMayMatch returns true if the parser should parse all files because at least one file requires it.
+func (p *Parser) aspectMayMatch() bool {
```
RomainMuller, line 122: The name appears to have little to do with the function's semantic value? It looks like this returns whether the file should be eagerly parsed or not? If so - why not name it `should[Eagerly]Parse`

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+			if p.aspectMayMatch() {
+				fileAspects = p.fileFilterAspects(fileAspects, p.rawFiles[idx])
+				if len(fileAspects) == 0 {
+					// No aspects can match on this file, no need to fill up the File.AstFile field.
+					p.parsedFiles[idx] = File{Name: file}
+					return
+				}
+			}
+
+			p.eagerness.Store(false)
```
RomainMuller, line 98: So we reset `eagerness` to `false` whenever we find the 1st file with aspects... That appears to be the opposite of what I'd have understood from the documentation... The semantics of this is fairly confusing at this stage...

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+			p.parsedFiles[idx] = p.parseFile(p.rawFiles[idx], fileAspects)
+		}(idx, file)
+	}
+
+	p.wg.Wait()
+
+	// No aspects can match on this package, return nothing
+	if !p.aspectMayMatch() {
+		return nil, nil
```
RomainMuller, line 107: Maybe there should be a dedicated boolean `hasApplicableAspects` or something

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	if len(p.errs) > 0 {
+		return nil, errors.Join(p.errs...)
```
RomainMuller, line 111: Here, if we have matched 0 aspects; but have encountered at least 1 error (perhaps preventing us from determining we'd have needed to apply some aspect), we silently ignore all the errors?

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// If we arrived here, this means we need to parse all files anyway because the type-checking pass will need them.
+	p.parseMissingFiles()
+	p.wg.Wait()
```
RomainMuller, line 116: Pretty awkward to have `p.wg.Wait()` done here, when the code in this function has already awaited the group... If anything new was added it was by `p.parseMissingFiles()` -- and so that function should have awaited before returning...?

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+
+	return p.parsedFiles, errors.Join(p.errs...)
+}
+
+// aspectMayMatch returns true if the parser should parse all files because at least one file requires it.
+func (p *Parser) aspectMayMatch() bool {
+	return !p.eagerness.Load() || p.filesBytesCount.Load() > maxBytesEagerness
+}
+
+func (p *Parser) addError(err error) {
```
RomainMuller, line 126: 1. All usage of this actually pass the result of an `fmt.Errorf` call, so you could have the signature here be:
   ```suggestion
   func (p *Parser) addError(format string, args... any) {
   ```
2. If you use `errgroup.Group` as suggested previously, I think this becomes redundant.

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+	p.errsMu.Lock()
+	defer p.errsMu.Unlock()
+	p.errs = append(p.errs, err)
+}
+
+func (p *Parser) parseFile(rawFile rawFile, aspects []*aspect.Aspect) File {
+	astFile, err := goparser.ParseFile(p.fset, rawFile.mappedName, rawFile.content, goparser.ParseComments)
+	if err != nil {
+		p.addError(fmt.Errorf("parsing %q: %w", rawFile.name, err))
+		return File{}
```
RomainMuller, line 136: I find this uncanny; and tracking the error list as state on the `*Parser` looks odd... It feels like this could return `(File, error)` and the error tracking could be done in a local variable on the caller side...

But again, `errgroup.Group` might actually naturally drive this.

internal/injector/parse/parser.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (p *Parser) parseMissingFiles() {
+	for i := range p.parsedFiles {
+		// Skip files that have already been parsed.
+		if p.parsedFiles[i].AstFile != nil {
+			continue
+		}
+
+		p.wg.Add(1)
```
RomainMuller, line 149: As mentioned earlier, this should mean the function needs to `p.wg.Wait()` before returning.

internal/injector/testdata/injector/http-no-arg-name/config.yml
```diff
@@ -37,6 +37,9 @@ aspects:
               {{ range .DirectiveArgs "dd:span" -}}{{printf "%q, %q," .Key .Value}}{{- end }}
             )
 
+import-map:
+  net/http: ""
```
RomainMuller, line 41: This also imports `log`... TBH I wonder if instead of specifying this by hand it could/should be grabbed from the AST as part of the test (parse for imports, then just harvest from the import specs).

internal/injector/testdata/injector/import-shadowing/expected.diff
RomainMuller, line 1: Is this change expected? I feel like this is effectively a regression?

internal/toolexec/aspect/oncompile.go
```diff
@@ -125,19 +125,19 @@ func (w Weaver) OnCompile(cmd *proxy.CompileCommand) (err error) {
 	}
 
 	injector := injector.Injector{
-		Aspects:    aspects,
 		RootConfig: map[string]string{"httpmode": "wrap"},
 		Lookup:     imports.Lookup,
 		ImportPath: w.ImportPath,
 		TestMain:   cmd.TestMain() && strings.HasSuffix(w.ImportPath, ".test"),
+		ImportMap:  imports.PackageFile,
```
RomainMuller, line 132: I feel like in this context this should be inferred from the source file instead of manually configured.

RomainMuller (changes requested):

internal/injector/aspect/may/context.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (ctx *PackageContext) PackageImports(path string) MatchType {
+	if path == "" {
+		return Unknown
+	}
+	_, ok := ctx.ImportMap[path]
+	if ok || path == ctx.ImportPath {
+		return Match
+	}
```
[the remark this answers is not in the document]
eliottness: I ordered statement in order of likely-ness. I believe this is the right way but if you have arguments in favor of not doing like this please tell me about them

internal/injector/aspect/join/configuration.go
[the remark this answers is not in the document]
eliottness: I believe this can be done in another PR (and I am not entirely okay with it haha)

internal/injector/check.go
eliottness: Please advise me how you would test this good sir

internal/injector/injector.go
eliottness: Since the return value was named `result` I figured this was in the theme lol.

internal/injector/injector_test.go
eliottness: You are right this is a set so a `map[string]struct{}` could be enough. But since a `map[string]string` is already filled for other purposes earlier in the code, doing one less copy of the map seemed more compelling.

internal/toolexec/aspect/oncompile.go
eliottness: WDYM ?

internal/injector/check.go
RomainMuller: Probably just create a fake package that declares `go 1.{runtime.VERSION + 1}`, and try to build it w/ `GOTOOLCHAIN=local`, etc...?

internal/injector/injector.go
RomainMuller: Touché!

internal/injector/injector_test.go
RomainMuller: This is the YAML for specifying test expectations... I feel like "comfort of specifying tests" prevails here...

internal/injector/builtin/integration_test.go
```diff
@@ -69,21 +70,33 @@ func Test(t *testing.T) {
 				return os.Open(file)
 			}
 
+			importMap := map[string]string{}
+			for _, a := range aspects {
+				for _, i := range a.JoinPoint.ImpliesImported() {
+					importMap[i] = ""
+				}
+			}
```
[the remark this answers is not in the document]
eliottness: This is used a the `Injector.ImportMap` value which is a map[string]string

internal/injector/parse/parser.go
eliottness: WDYM ?

internal/injector/testdata/injector/import-shadowing/expected.diff
eliottness: Merge conflict, fixed

internal/injector/check_test.go
```diff
@@ -0,0 +1,44 @@
...
+	astFile, err := goparser.ParseFile(fset, "main.go", []byte(goFile), goparser.ParseComments)
+	require.NoError(t, err)
+
+	versionInt, err := strconv.Atoi(runtime.Version()[4:6])
+	require.NoError(t, err)
+
+	injector := &Injector{
+		GoVersion: fmt.Sprintf("go1.%d", versionInt+1), // +1 to always make sure we have a newer version
+		Lookup: func(_ string) (io.ReadCloser, error) {
+			return nil, nil
```
RomainMuller, line 38: Just to make sure this does not cause a nil-deref that is confusing to diagnose... ^^

```suggestion
			return nil, errors.ErrUnsupported
```

RomainMuller (approved):
SAAC

Issue #445, opened 2024-12-03
Missing: integration test for `net/http.DefaultHandler`


RomainMuller:
Fix introduced in https://github.com/DataDog/orchestrion/pull/440 but not yet covered by integration tests.

PR #447, opened 2024-12-05
test: integration test for `net/http.DefaultServeMux`

keisku:
Fixes https://github.com/DataDog/orchestrion/issues/445

### DIFF
--- a/_integration-tests/tests/net_http/serve_mux_handler.go
+++ b/_integration-tests/tests/net_http/serve_mux_handler.go
@@ -8,6 +8,7 @@
 package nethttp
 
 import (
+	"net/http"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


_integration-tests/tests/net_http/serve_mux_handler.go
```diff
@@ -19,3 +20,14 @@ func (tc *TestCaseServeMuxHandler) Setup(t *testing.T) {
...
 	tc.base.Setup(t)
 }
+
+type TestCaseHandlerIsNil struct {
+	base
+}
+
+func (tc *TestCaseHandlerIsNil) Setup(t *testing.T) {
+	http.HandleFunc("/hit", tc.base.handleHit)
+	http.HandleFunc("/", tc.base.handleRoot)
```
RomainMuller, line 30: This could introduce temporal coupling... I wonder if this should restore the default handler afterwards... But I'm also not entirely convinced this is actually any problem in this context... so just nothing this here for posterity in case it turns out it was and we stumble on this PR trying to figure out what is going on...

RomainMuller (approved):

PR #456, opened 2024-12-06
refactor(config): use different jsonschema dependency at runtime

eliottness:
Because the other one was making a truckload of allocations at init


_integration-tests/grpc.test
RomainMuller, line 1: I suspect you might not have wanted to check this one in...

eliottness: Fixed

PR #478, opened 2024-12-19
chore: switch logging framework to zerolog

RomainMuller:


internal/ensure/requiredversion.go
```diff
@@ -128,7 +132,7 @@ func requiredVersion(
 // required in the specified directory's "go.mod" file. If dir is blank, the process' current
 // working directory is used. The version may be blank if a replace directive is in effect; in which
 // case the path value may indicate the location of the source code that is being used instead.
-func goModVersion(dir string) (moduleVersion string, moduleDir string, err error) {
+func goModVersion(log *zerolog.Logger, dir string) (moduleVersion string, moduleDir string, err error) {
```
eliottness, line 135: I believe sending the logger as arguments is against recommendations. Why not send the context and run `zerolog.Ctx` locally ?

internal/jobserver/pkgs/resolve.go
```diff
@@ -132,10 +138,12 @@ func (s *service) resolve(req *ResolveRequest) (ResolveResponse, error) {
 			}
 			env = append(env, fmt.Sprintf("%s=%s", envVarGotmpdir, req.TempDir))
 		}
+		// Artifically limit to 2 in order to reduce scheduler contention in downstream processes...
+		env = append(env, "GOMAXPROX=2")
```
eliottness, line 142: sneaky one. Not sure it should be here and not sure we should do it like this. Let's do another PR of this no ?

eliottness (changes requested):

eliottness (approved):

PR #479, opened 2024-12-20
chore: exponential backoff in flaky kafka test

RomainMuller:
The segmentio_kafka test has been falking due to the test container not having the topic and partition ready by the time the test tries to produce messages. This is possiblye due to the test waiting up to 1 second, while the container has been observed to need almost 2 seconds to transition the initial partition to available.

### DIFF
--- a/_integration-tests/utils/backoff/backoff.go
+++ b/_integration-tests/utils/backoff/backoff.go
@@ -66,5 +66,5 @@ func doRetry(
 		}
 	}
 
-	return errs
+	return errors.Join(errs, ctx.Err())
 }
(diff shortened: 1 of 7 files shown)


darccio (approved):

_integration-tests/utils/backoff/backoff.go
```diff
@@ -0,0 +1,105 @@
...
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+// Package backoff provides utilities to retry operations when encoutering
+// transient errors. It is used by integration tests to allow for testcontainers
+// enough time to become fully ready, avoiding tests flaking because the CI
+// resources are constrained enough to cause containers to not be "immediately"
+// ready to serve traffic.
+package backoff
```
rarguelloF, line 11: I wonder if we could use https://github.com/cenkalti/backoff instead of implementing this ourselves 🤔

rarguelloF (commented):
LGTM! just wondering if we can use an existing library (https://github.com/cenkalti/backoff) for the exponential backoff retries

PR #489, opened 2025-01-06
feat(injector: join): signature-contains

eliottness:
Adds flexible function signature matching to support partial and order-independent type matching. 

This enables simpler instrumentation configuration when targeting functions that share common types (e.g. context.Context or error) regardless of their position or other parameters. 

The change simplifies instrumentation configuration by reducing the need for multiple matchers when targeting related function groups

## Changes

- Added SignatureContains matcher that checks for presence of types anywhere in function signatures
- Supports matching on partial argument/return types without requiring exact matches
- Added tests

### DIFF
--- a/internal/injector/aspect/join/function.go
+++ b/internal/injector/aspect/join/function.go
@@ -217,6 +217,40 @@ func (fo *signature) Hash(h *fingerprint.Hasher) error {
 	)
 }
 
+type signatureContains struct {
+	signature
+}
+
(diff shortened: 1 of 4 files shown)


kakkoyun (commented):
What do we need to push this through the finish line?

eliottness:
@kakkoyun We would need to write the corresponding tests and that would be it

darccio (approved):

PR #490, opened 2025-01-07
feat: never build twice

### LINKED ISSUE #430: [BUG] Packages may be (unnecessarily) built multiple times
**Version of orchestrion**
1.0.0

**Describe what happened:**
Packages that are added by injection (including their transitive closure) may be built multiple times:
- Several resolutions of the same transitive dependency happening in parallel on a cold cache
- A "natural" dependency of the instrumented project that is shared with injected packages (that is a cache miss) will always be built as part of the normal build's flow

The job server's build graph could be enhanced to include more information and allow complete de-duplication of individual package compilation outputs, so that any given package is only built exactly once.

This would result in lower concurrency count, which likely will improve overall build performance.

RomainMuller:
Adds a new jobserver message to track ongoing compilation tasks, so that duplicates can be completely avoided by re-using the first build task's outcome. This allows de-duplicating builds even if an injected package is already present in the natural dependency closure of the package being built.

Fixes #430

### DIFF
--- a/internal/jobserver/nbt/nbt.go
+++ b/internal/jobserver/nbt/nbt.go
@@ -117,8 +117,8 @@ const (
 	LabelAsmhdr  Label = "go_asm.h"
 )
 
-func (StartRequest) Subject() string             { return startSubject }
-func (*StartResponse) IsResponseTo(StartRequest) {}
+func (StartRequest) Subject() string           { return startSubject }
+func (StartRequest) ResponseIs(*StartResponse) {}
 
 func (s *service) start(ctx context.Context, req StartRequest) (*StartResponse, error) {
 	if req.ImportPath == "" || req.BuildID == "" {
@@ -183,8 +183,8 @@ type (
 	}
 )
 
-func (FinishRequest) Subject() string              { return finishSubject }
-func (*FinishResponse) IsResponseTo(FinishRequest) {}
+func (FinishRequest) Subject() string            { return finishSubject }
+func (FinishRequest) ResponseIs(*FinishResponse) {}
 
 var errNoFilesNorError = errors.New("missing files, and no error reported")
 
@@ -214,7 +214,7 @@ func (s *service) finish(ctx context.Context, req FinishRequest) (*FinishRespons
 
 	if !state.isDone.CompareAndSwap(false, true) {
 		log.Info().Msg("Task was already completed (concurrent retry?)")
-		return &FinishResponse{}, nil
+		return nil, nil
 	}
 
 	defer state.onDone()
--- a/internal/jobserver/nbt/nbt.go
+++ b/internal/jobserver/nbt/nbt.go
@@ -35,6 +35,7 @@ type (
 	}
 	buildState struct {
 		initOnce sync.Once
+		buildID  string
 		token    string          // Finalization token
 		onDone   func()          // Called once the original task has completed
 		done     <-chan struct{} // Blocks until the original task has completed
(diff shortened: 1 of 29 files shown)


internal/jobserver/nbt/nbt.go
```diff
@@ -0,0 +1,272 @@
...
+		state.additionalFiles = nil
+		state.archive = ""
+		state.error = fmt.Errorf("persisting archive %q: %w", req.ArchivePath, err)
+		return nil, state.error
+	}
+
+	return &FinishResponse{}, nil
+}
+
+var ns = uuid.MustParse("4BFB6F4B-212C-43A0-A581-A29C8B3D3BE4")
```
eliottness, line 272: :broom:

internal/jobserver/nbt/nbt.go
```diff
@@ -0,0 +1,272 @@
...
+		FinishToken string `json:"token"`
+		// ArchivePath is the path to the archive produced by the compilation task.
+		// It must not be blank unless [FinishRequest.Error] is not nil. If present,
+		// the file must exist on disk, and will be hard-linked (or copied) into a
+		// temporary directory so it can be re-used by subsequent [StartRequest] for
+		// the same [FinishRequest.ImportPath].
+		ArchivePath string `json:"archivePath,omitempty"`
+		// AdditionalFiles is a list of additional files produced by the compilation
+		// task, associated to a user-defined label.
+		AdditionalFiles map[Label]string `json:"extra,omitempty"`
```
eliottness, line 183: I did not finish the review yet but I don't see any places where `ArchivePath` receives a different treatment from `AdditionalFiles`. So maybe we could just merge it into a `Files map[Label]string` attribute.

internal/toolexec/aspect/oncompile.go
```diff
@@ -65,18 +73,85 @@ var weavingSpecialCase = []specialCase{
...
+	log := zerolog.Ctx(ctx).With().Str("phase", "compile").Str("import-path", w.ImportPath).Logger()
 	ctx = log.WithContext(ctx)
 
-	imports, err := importcfg.ParseFile(cmd.Flags.ImportCfg)
-	if err != nil {
-		return fmt.Errorf("parsing %q: %w", cmd.Flags.ImportCfg, err)
+	if js, err := client.FromEnvironment(ctx, cmd.WorkDir); err != nil {
+		log.Debug().Str("work-dir", cmd.WorkDir).Err(err).Msg("Failed to obtain job server client")
+	} else {
+		res, err := client.Request[nbt.StartRequest, *nbt.StartResponse](ctx, js, nbt.StartRequest{ImportPath: w.ImportPath, BuildID: cmd.Flags.BuildID})
```
eliottness, line 83: Is it normal that one type argument is a struct and the other is a pointer to a struct? Was specifing the type arguments necessary in the first place? I thought the `IsResponseTo` function was a solution for that

internal/toolexec/aspect/oncompile.go
```diff
@@ -65,18 +73,85 @@ var weavingSpecialCase = []specialCase{
...
+			_, err := client.Request[nbt.FinishRequest, *nbt.FinishResponse](ctx, js, nbt.FinishRequest{
+				ImportPath:      w.ImportPath,
+				FinishToken:     res.FinishToken,
+				ArchivePath:     archivePath,
+				AdditionalFiles: additionalFiles,
+				Error:           error,
+			})
+			return err
+		})
+	}
```
eliottness, line 145: I believe all of this could get it's own function (more than one would be ideal of course)

eliottness: FYI `OnCompile` is about 225 lines long :eye: :eye:

eliottness (changes requested):

internal/toolexec/aspect/oncompile.go
RomainMuller: `IsResponseTo` prevents mis-use, but Go generic inference is too weak to successfully handle this >_>

RomainMuller: Reversing the generics (request tells what its response is, instead of response tells what its request is) allows the inference to work :)

eliottness (approved):

PR #493, opened 2025-01-09
feat: move integrations configuration to dd-trace-go

RomainMuller:


internal/injector/builtin/all.yml
RomainMuller, line 1: This is not use -- get rid.

eliottness (approved):

Issue #511, opened 2025-01-27
[BUG] Auto instrumentation with gorm fails


AlonGluz:
**Version of orchestrion**
1.0.3

**Describe what happened:**
not enough arguments in call to __orchestrion_gorm.Open
	have ("gorm.io/gorm".Dialector)
	want ("gorm.io/gorm".Dialector, *"gorm.io/gorm".Config, ..."gopkg.in/DataDog/dd-trace-go.v1/contrib/gorm.io/gorm.v1".Option)
exit status 2

**Describe what you expected:**
For it to work, gorm changed the function interface and  *"gorm.io/gorm".Config required.
**Steps to reproduce the issue:**
build go microservice that utilizes gorm 

**Additional environment details (Version of Go, Operating System, etc.):**
go 1.23
gorm 1.25.12

RomainMuller:
Hey!

Thanks for reporting this... Indeed, the standard `gorm` API does not demand a `gorm.Config` first argument... but the tracer API we replace it with does... This is an oversight from us.

AlonGluz:
So the bug should be fixed when you will merge #512 ? 
Appreciate the quick response, I'll   //orchestrion:ignore it for now, until you merge it

RomainMuller:
This should be fixed in `v1.0.4` :) Please let us know if that is not the case!

PR #512, opened 2025-01-27
Fix gorm.io builtin aspect

rarguelloF:
Fixes https://github.com/DataDog/orchestrion/issues/511

`gorm.io/gorm` aspect: Do not replace with a function with a different signature.


RomainMuller:
Released as part of `v1.0.4`.

Issue #491, opened 2025-01-08
go.mod changed during editing; not overwriting


eliottness:
**Version of orchestrion**

`v1.0.2`

Dockerfile:
```
RUN mkdir -p /usr/local/bin 

RUN GOBIN=/usr/local/bin go install github.com/DataDog/orchestrion@latest

# Build the Go binary
ENV GOFLAGS="${GOFLAGS} '-toolexec=/usr/local/bin/orchestrion toolexec'"
```

**Describe what happened:**

```
#13 [8/8] RUN go build -o main .
#13 0.255 Version check error: failed to determine go.mod requirement for "github.com/DataDog/orchestrion": -: no required module provides package github.com/DataDog/orchestrion; to add it:
#13 0.255 	go get github.com/DataDog/orchestrion
#13 0.255 Warning: github.com/DataDog/orchestrion is not present in your go.mod file.
#13 0.255 In order to ensure build reliability and reproductibility, orchestrion will now add itself in your go.mod file by:
#13 0.255 
#13 0.255     1. creating a new file named orchestrion.tool.go
#13 0.255     2. running go get github.com/DataDog/orchestrion@v1.0.2
#13 0.255     3. running go mod tidy
#13 0.255 
#13 0.255 You should commit the resulting changes into your source control system.
#13 0.255 go: go.mod changed during editing; not overwriting
#13 0.255 Failed to pin orchestrion in go.mod: editing "/app/go.mod": running `go mod edit [-require=github.com/DataDog/orchestrion@v1.0.2]`: exit status 1
#13 0.255 go: error obtaining buildID for go tool compile: exit status 1
#13 ERROR: process "/bin/sh -c go build -o main ." did not complete successfully: exit code: 1
------
 > [8/8] RUN go build -o main .:
0.255 In order to ensure build reliability and reproductibility, orchestrion will now add itself in your go.mod file by:
0.255 
0.255     1. creating a new file named orchestrion.tool.go
0.255     2. running go get github.com/DataDog/orchestrion@v1.0.2
0.255     3. running go mod tidy
0.255 
0.255 You should commit the resulting changes into your source control system.
0.255 go: go.mod changed during editing; not overwriting
0.255 Failed to pin orchestrion in go.mod: editing "/app/go.mod": running `go mod edit [-require=github.com/DataDog/orchestrion@v1.0.2]`: exit status 1
0.255 go: error obtaining buildID for go tool compile: exit status 1
------
Dockerfile:30
--------------------
[...]

PR #521, opened 2025-01-31
fix: auto-pin race condition on go.mod file

RomainMuller:
When using `-toolexec` mode, the auto-pin process may be initiated by multiple concurrent executions of the `toolexec` proxy. In such cases, there is a race condition on access to `go.mod` and `orchestrion.tool.go`, which often results in `go mod tidy` failing due to it detecting concurrent modifications of the `go.mod` (it may also result in corrupt content of the `orchestrion.tool.go` file).

This addresses the issue by having the auto-pin process acquire an advisory write lock on the `go.mod` file, making sure all attempts are properly synchronized, and only one attempt tries to modify the file at a time.

Fixes https://github.com/DataDog/orchestrion/issues/491


kakkoyun (approved):
LGTM

PR #525, opened 2025-02-04
Upgrade to dd-trace-go v1.71.1

mtoffl01:
Upgrades orchestrion to use new dd-trace-go patch release v1.71.1 and increases orchestrion's internal version tag.


internal/version/version.go
```diff
@@ -9,7 +9,7 @@ import "runtime/debug"
 
 const (
 	// tag specifies the current release tag. It needs to be manually updated.
-	tag       = "v1.0.3"
+	tag       = "v1.0.4"
```
mtoffl01, line 12: Was not sure what the correct value should be. Is this a patch release or minor release for orchestrion? Is the answer determined by whether we're upgrading minor or patch versions of dd-trace-go (in this case, the latter)?

Please clarify and I can update instructions in the template: https://datadoghq.atlassian.net/wiki/spaces/DL/pages/2477949158/Template+v+MAJOR+.+MINOR+.+PATCH+Release+Checklist#Validate-the-release

RomainMuller: Usually we do a patch release with these.

RomainMuller:
Superceded by `v1.1.0-rc.1` in #533

PR #534, opened 2025-02-07
fix: windows command lines need parsing with `CommandLineToArgvW`

RomainMuller:
Windows command line arguments are conveyed as a string, which needs to be processed in order to obtain the "accurate" string slice with quoting removed and proper spacing handling. The `windows.CommandLineToArgv` function provides support for this.

---

This prompted enabling Windows unit test execution, which uncovered a number of issues (some real Windows bugs, some test harness oversights), which this PR also addresses.

---

The job server startup process also does not provide enough log detail to determine the cause of issues on certain Windows CI systems (possible cause being they're just extremely slow). This adds more logging to better identify what is going on.


internal/jobserver/client/env.go
```diff
@@ -120,25 +135,74 @@ func clientFromURLFile(ctx context.Context, path string) (*Client, string, error
...
+
+func init() {
+	val := os.Getenv("ORCHESTRION_JOB_SERVER_START_TIMEOUT_SECONDS")
+	if val == "" {
+		return
+	}
+
+	sec, err := strconv.Atoi(val)
+	if err != nil {
+		// We got an invalid value, we'll silently ignore this because we can't ensure the logger has been initialized yet.
```
eliottness, line 203: This is a very light cause to not log an error that could make us gain a lot of time in the future in case of typo in this env var

eliottness (approved):

tonyredondo (approved):

PR #540, opened 2025-02-10
fix: pragma directives not picked up on return statement

RomainMuller:
The `//dd:span` directive is not honored when it's presented on a `return` statement (where the expression is a literal function expression).

This changes the directive lookup function so it crawls up to the parent if it's a node type that wraps a single expression; effectively:
- `name := <expr>` (LHS and RHS)
- `name = <expr>` (RHS only)
- `<call>` (function portion only)
- `defer <call>`
- `<expr>` (the expression statement)
- `go <expr>`
- `return <expr>`
- `ch <- <expr>` (the value only)

Fixes #539

### DIFF
--- a/internal/injector/aspect/join/directive.go
+++ b/internal/injector/aspect/join/directive.go
@@ -49,10 +49,14 @@ func (d directive) matchesChain(chain *context.NodeChain) bool {
 		}
 	}
 
-	// If the parent is an assignment statement, so we also check it for directives.
 	if parent := chain.Parent(); parent != nil {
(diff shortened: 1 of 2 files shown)


internal/injector/aspect/join/directive.go
```diff
@@ -49,10 +60,28 @@ func (d directive) matchesChain(chain *context.NodeChain) bool {
 		}
 	}
 
-	// If the parent is an assignment statement, so we also check it for directives.
 	if parent := chain.Parent(); parent != nil {
```
eliottness, line 63:
```suggestion
	parent := chain.Parent()
	if parent == nil {
		return false
	}
```

eliottness (approved):

PR #545, opened 2025-02-17
fix: job server fails to start on Windows due to locking error

RomainMuller:
Updated the `*filelock.Mutex` type to implement `io.ReadWriteSeeker`, `io.ReaderAt`, and `io.WriterAt` so that it can be used to read/write the locked file without having to open a secondary file descriptor. This is necessary on Windows, where the locks are tied to the specific file descriptor (as opposed to UNIX where the lock is tied to the process).


internal/filelock/filelock.go
```diff
@@ -21,6 +24,17 @@ import (
 //
 // Upgrading a read-lock to a write lock, or vice-versa, is not guaranteed to
 // happen atomically (on Windows, it is guaranteed not to be atomic).
+//
+// The [io.ReadWriteSeeker], [io.ReaderAt], and [io.WriterAt] interfaces are
+// implemented by [*Mutex] and are saef to be used once any lock as been
```
eliottness, line 29:
```suggestion
// implemented by [*Mutex] and are safe to be used once any lock as been
```

eliottness (approved):

PR #551, opened 2025-02-21
chore(docs): Fix typos and spellcheck

kakkoyun:
Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>


.github/workflows/validate.yml
```diff
@@ -150,6 +150,12 @@ jobs:
         run: go run ./_tools/headercheck/header_check.go
       - name: vet
         run: go vet ./...
+      - name: Check documentation for misspellings
+        uses: crate-ci/typos@v1.29.9
```
kakkoyun, line 154: More accurate then misspell https://github.com/crate-ci/typos

_docs/content/docs/architecture.md
```diff
@@ -215,7 +215,7 @@ Finally, it invokes the `go tool link` with updated arguments (⑧).
 Orchestrion drives code injection using a process similar to classical
 Aspect-oriented Programming (AoP) (see [Aspects][contrib-aspects]). These
 combine a _Join Point_ (where code needs to be modified) with one or more
-_Advices_ (what modifications need to be made).
+_Advice_s (what modifications need to be made).
```
eliottness, line 218: Is this one really something ? Is the italicizing working at all in this case

eliottness (commented):
Nice addition 👌

_docs/content/docs/architecture.md
RomainMuller: I don't think it does, at least not in commonmark standard, which says it needs to be followed by a unicode white space code point for it to "count".

kakkoyun: Good catch. I didn't check how it got rendered. The error is also a bit weird:
```text
error: `Advices` should be `Advice`, `Advises`
  --> _docs/content/docs/architecture.md:218:2
    |
218 | _Advices_ (what modifications need to be made).
```

I will add a config to exclude this one. There isn't a cleaner way to render otherwise.

RomainMuller:
Well turns out "advice" is a mass word in English (meaning it has no plural). So the correct fix is to de-pluralize it.

Issue #560, opened 2025-02-28
[BUG] Can not build amd64 image on ARM Mac


zedimir:
Hi there,

I've been trying to instrument my app with orchestrion. While it worked flawless on linux, I couldn't build an amd64 image on mac

Also, thank you for the tool. It's cool!

**Version of orchestrion**
orchestrion v1.1.0

**Describe what happened:**
Depending on the container tool (Docker Desktop and OrbStack) and virtualization they use, the errors are different but all could be boiled down to `rosetta/qemu not found`.

More details here: https://github.com/3timeslazy/dd-orchestrion-rosetta/tree/main

I was able to fix the problem with "rosetta not found" (see the repo's readme), however the build still fails (last section in the repo's readme)

Also, might be related to https://github.com/DataDog/orchestrion/issues/358

**Additional environment details (Version of Go, Operating System, etc.):**

Mac: Apple M3 Max
OrbStack: Version 1.10.2 (19048)
Docker Desktop: 4.38.0 (181591)
Go: 1.23

RomainMuller:
Alright it seems like the emulation layer discover isn't quite nailed out just yet -- thanks for reporting... I think your fix looks acceptable, so feel free to open a PR with it.

---

The log at the end of your repo, for the failure after the fix, ends with:
```
dd-orchestrion-rosetta: go build dd-orchestrion-rosetta: copying /tmp/go-build444885451/b001/exe/a.out: open main: read-only file system
````

This looks like this issue would be that `$TMPDIR` is on a read-only file system? You probably want to check this... I would expect A LOT of things fail if the temporary directory is read-only.

RomainMuller:
This has been hopefully fixed by v1.2.0... Feel free to re-open if there is still an issue!

PR #565, opened 2025-03-05
fix: rosetta not being detected when command is [rosetta]

3timeslazy:
Fix for the "rosetta not found" error from here: https://github.com/DataDog/orchestrion/issues/560

### DIFF
--- a/internal/goflags/flags.go
+++ b/internal/goflags/flags.go
@@ -399,7 +399,9 @@ func parentGoCommandFlags(ctx context.Context, pid int) (flags CommandFlags, err
 		// the registered rosetta binfmt handler. In such cases, the argv0 has a leaf name of "rosetta"
 		// and is not present within the container itself (it's only on the hypervisor). In such cases,
 		// we try to resolve argv[1] instead. This can only manifest itself on amd64 + linux.
-		if errors.Is(err, fs.ErrNotExist) && runtime.GOARCH == "amd64" && runtime.GOOS == "linux" && filepath.Base(args[0]) == "rosetta" && len(args) > 1 {
+		notExist := errors.Is(err, fs.ErrNotExist) || strings.Contains(err.Error(), "executable file not found")
+		rosetta := strings.Contains(filepath.Base(args[0]), "rosetta")
+		if notExist && runtime.GOARCH == "amd64" && runtime.GOOS == "linux" && rosetta && len(args) > 1 {
 			log.Trace().Err(err).Msg("Attempting to resolve rosetta target after error resolving argv0")
 			var err2 error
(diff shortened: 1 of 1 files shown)


3timeslazy:
In the PR I've only added rosetta check, but if the maintainers would like, I could add qemu check as well

RomainMuller:
> In the PR I've only added rosetta check, but if the maintainers would like, I could add qemu check as well

That would be super helpful thank you <3

3timeslazy:
Added the qemu check. Also, added the `err != nil` check, because if everything ok and the err is nil, `errors.Is` will return false, while the new statement will panic. Fix that

internal/goflags/flags.go
```diff
@@ -399,7 +399,10 @@ func parentGoCommandFlags(ctx context.Context, pid int) (flags CommandFlags, err
 		// the registered rosetta binfmt handler. In such cases, the argv0 has a leaf name of "rosetta"
 		// and is not present within the container itself (it's only on the hypervisor). In such cases,
 		// we try to resolve argv[1] instead. This can only manifest itself on amd64 + linux.
-		if errors.Is(err, fs.ErrNotExist) && runtime.GOARCH == "amd64" && runtime.GOOS == "linux" && filepath.Base(args[0]) == "rosetta" && len(args) > 1 {
+		notExist := (err != nil && (errors.Is(err, fs.ErrNotExist) || strings.Contains(err.Error(), "executable file not found")))
+		emulation := strings.Contains(filepath.Base(args[0]), "rosetta")
+		emulation = emulation || strings.Contains(filepath.Base(args[0]), "qemu")
```
RomainMuller, line 404:
```suggestion
		base := filepath.Base(args[0])
		emulation := strings.Contains(base, "rosetta") || strings.Contains(base, "qemu")
```

internal/goflags/flags.go
```diff
@@ -399,7 +399,10 @@ func parentGoCommandFlags(ctx context.Context, pid int) (flags CommandFlags, err
...
+		if notExist && runtime.GOARCH == "amd64" && runtime.GOOS == "linux" && emulation && len(args) > 1 {
```
RomainMuller, line 405: While the Rosetta check is only relevant for AMD64, I feel the QEMU one should be relevant for more architectures?

RomainMuller (commented):
A couple nitpicks but nothing major. Thanks for this!

internal/goflags/flags.go
RomainMuller: That's fine by me.

RomainMuller (approved):

PR #570, opened 2025-03-10
feat: new //orchestrion:version pragma

RomainMuller:
Introduce a new //orchestrion:version pragma to
allow run-time determination of the orchestrion
version that built a given binary.

This would be useful to implement telemetry directly
in the tracer so it is reliably sent even when the
tracer is not started by orchestrion itself.

### DIFF
--- a/internal/injector/config/builtin.go
+++ b/internal/injector/config/builtin.go
@@ -34,11 +34,26 @@ var builtIn = configGo{
 					),
 				},
 			},
+			{
+				ID: "built.WithOrchestrionVersion",
+				JoinPoint: join.AllOf(
+					join.ValueDeclaration(join.MustTypeName("string")),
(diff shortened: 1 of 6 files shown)


internal/injector/config/builtin.go
```diff
@@ -34,11 +34,26 @@ var builtIn = configGo{
 					),
 				},
 			},
+			{
+				ID: "built.WithOrchestrionVersion",
+				JoinPoint: join.AllOf(
```
eliottness, line 39:
```suggestion
				TracerInternal: true,
				JoinPoint: join.AllOf(
```

eliottness (changes requested):

eliottness (approved):

kakkoyun (approved):

PR #577, opened 2025-03-20
feat: instrument Orchestrion with Datadog tracing

RomainMuller:
In order to facilitate investigation of build performance, instrument Orchestrion (manually, because who does not love some irony) with the `dd-trace-go` V2 library, so that we can obtain traces from Orchestrion executions.

This is enabled by setting `ORCHESTRION_TRACE` to any non-blank value, and requires a Datadog Agent to be available and appropriately configured in order for traces to be sent to the backend.

Child process distributed tracing is achieved by using environment variables as a propagation medium, using a custom-made carrier that works off of `os.Environ()`-style slices.

The NATS client/server is also instrumented for distributed tracing, using `nats.Msg.Header` as a carrier (this is akin to using HTTP or gRPC headers for the same purpose).

------------------------------------------------------------------------

Initial investigations of traces (building orchestrion with orchestrion) showed that a far-from-trivial overhead was induced by an abundance of sub-process creation to `go tool pack` (involved in storing & accessing the `link.deps` files). Replacing these with a simple library (`github.com/blakesmith/ar`) allowed reducing the overhead of the `Weaver.OnLink` call by a significant amount (shaving 6 seconds off of this particular build).


.github/workflows/validate.yml
```diff
@@ -227,7 +227,12 @@ jobs:
   benchmark:
     needs: generate
     runs-on: arm-8core-linux
-    name: Benchmarks
+    name: Benchmarks (run ${{ matrix.run_number }})
+    strategy:
+      matrix:
+        # Run benchmarks 6 times on different runners to smooth out compute discrepancies...
+        run_number: [1, 2, 3, 4, 5, 6]
+      fail-fast: false
```
eliottness, line 235:
```suggestion
      fail-fast: true
```
Why not ?

.github/workflows/validate.yml
```diff
@@ -240,20 +245,41 @@ jobs:
...
+      - name: Setup go
+        id: setup-go
+        uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5
+        with:
+          go-version: stable
+          cache-dependency-path: "**/go.mod"
+      - name: Download benchmark reports (raw)
+        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4
+        with:
+          pattern: Benchmark Report (run *)
```
eliottness, line 274: doing globbing with with spaces in them sure is wierd. It took me sometime to understand
```suggestion
          name: benchmarks-${{ matrix.run_number }}.txt
          path: ${{ runner.temp }}/benchmarks-${{ matrix.run_number }}.txt

  benchmark-report:
    needs: benchmark
    runs-on: ubuntu-latest
    name: Benchmark Report
    steps:
      - name: Checkout
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
      - name: Setup go
        id: setup-go
        uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # v5
        with:
          go-version: stable
          cache-dependency-path: "**/go.mod"
      - name: Download benchmark reports (raw)
        uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4
        with:
          pattern: benchmarks-*.txt
```

internal/toolexec/aspect/linkdeps/linkdeps.go
```diff
@@ -202,31 +202,33 @@ func (l *LinkDeps) Write(w io.Writer) error {
...
-			break
+		if hdr.Name != entry {
+			continue
+		}
+		data := make([]byte, hdr.Size)
+		n, err := rd.Read(data)
+		if err != nil {
+			return nil, fmt.Errorf("reading entry from archive: %w", err)
 		}
+		return bytes.NewReader(data[:n]), nil
```
eliottness, line 228: Optional -- Why not directly wrap the ar reader into a io.LimitReader and in make the function return a io.ReadCloser. Or use io.SectionReader

internal/toolexec/proxy/compile.go
```diff
@@ -160,23 +161,37 @@ func (cmd *CompileCommand) attachLinkDeps(ctx gocontext.Context) error {
 		return fmt.Errorf("mkdir %q: %w", orchestrionDir, err)
 	}
 
-	linkDepsFile := filepath.Join(orchestrionDir, linkdeps.Filename)
-	if err := cmd.LinkDeps.WriteFile(linkDepsFile); err != nil {
-		return fmt.Errorf("writing %s file: %w", linkdeps.Filename, err)
+	var buf bytes.Buffer
+	if err := cmd.LinkDeps.Write(&buf); err != nil {
+		return fmt.Errorf("writing "+linkdeps.Filename+": %w", err)
```
eliottness, line 166: I am racking my brain to understand why you did that but nothing seems to match...

internal/toolexec/proxy/compile.go
```diff
@@ -160,23 +161,37 @@ func (cmd *CompileCommand) attachLinkDeps(ctx gocontext.Context) error {
...
+		return fmt.Errorf("opening archive: %w", err)
+	}
+	defer file.Close()
+
+	wr := ar.NewWriter(file)
+	if err := wr.WriteHeader(&ar.Header{Name: linkdeps.Filename, Mode: 0o644, Size: int64(buf.Len())}); err != nil {
+		return fmt.Errorf("writing header: %w", err)
+	}
+	if _, err := wr.Write(buf.Bytes()); err != nil {
+		return fmt.Errorf("writing "+linkdeps.Filename+" entry: %w", err)
```
eliottness: Optional -- You could add a method to `LinkDeps` to predict the size of the file and then directly do `LinkDeps.Write(wr)`

main.go
```diff
@@ -37,6 +40,32 @@ const (
...
+			opts = append(opts, tracer.ChildOf(spanCtx))
+		}
+		opts = append(opts,
+			tracer.Tag("process.args", os.Args[1:]),
+			tracer.Tag("process.pid", os.Getpid()),
+		)
+
+		var span *tracer.Span
+		span, ctx = tracer.StartSpanFromContext(ctx, "main", opts...)
+		defer span.Finish()
```
eliottness: This main function is big enough maybe put this at least in another function and return just the span and the context

internal/injector/config/config-go.go
```diff
@@ -80,8 +80,14 @@ func (l *Loader) loadGoFile(ctx context.Context, filename string) (_ []Config, e
...
 
-	span, ctx := tracer.StartSpanFromContext(ctx, "config.loadGoFile", tracer.Tag("filename", filename))
-	defer func() { span.Finish(tracer.WithError(err)) }()
+	span, ctx := tracer.StartSpanFromContext(ctx, "config.loadGoFile", tracer.ResourceName(filename))
+	defer func() {
+		if !errors.Is(err, fs.ErrNotExist) {
+			span.Finish(tracer.WithError(err))
+		} else {
+			span.Finish()
+		}
```
eliottness, line 89:
```suggestion
		spanErr := err
		if errors.Is(err, fs.ErrNotExist) {
			spanErr = nil
		}
		span.Finish(tracer.WithError(err))
```

internal/injector/config/config-yml.go
```diff
@@ -34,7 +34,13 @@ func (l *Loader) loadYMLFile(ctx context.Context, dir string, name string) (_ *c
 	span, ctx := tracer.StartSpanFromContext(ctx, "config.loadYMLFile",
 		tracer.ResourceName(filename),
 	)
-	defer func() { span.Finish(tracer.WithError(err)) }()
+	defer func() {
+		if !errors.Is(err, fs.ErrNotExist) {
+			span.Finish(tracer.WithError(err))
+		} else {
+			span.Finish()
+		}
```
eliottness, line 42:
```suggestion
		spanErr := err
		if errors.Is(err, fs.ErrNotExist) {
			spanErr = nil
		}
		span.Finish(tracer.WithError(err))
```

internal/jobserver/pkgs/load.go
```diff
@@ -0,0 +1,90 @@
...
+	// LoadRequest is a request to load packages relative to a specific directory. It only loads the
+	// packages' names and (source) files, not their dependencies or export file. The result is cached
+	// for a given Dir+Pattern pair. Each pattern is loaded individually (so that they can be cached
+	// independently).
+	LoadRequest struct {
+		Dir      string   `json:"dir"`      // The directory to resolve from (usually where `go.mod` is)
+		Patterns []string `json:"patterns"` // Package pattern to resolve
+	}
+	// LoadResponse is the response to a [LoadRequest]. It contains the packages that were loaded.
+	LoadResponse []*packages.Package
```
eliottness, line 28: Are we sure there is no edge case where serializing packages.Package won't break ? Could we cleanse out fields we don't need or do a partial copy of the fields in the struct ?

internal/injector/config/config.go
```diff
@@ -26,16 +26,19 @@ type Config interface {
 	visit(Visitor, string) error
 }
 
+type PackageLoader = func(context.Context, string, ...string) ([]*packages.Package, error)
```
eliottness, line 29:
```suggestion
type PackageLoader func(context.Context, string, ...string) ([]*packages.Package, error)
```
Any reason to use the equal form here ?

internal/injector/config/config.go
```diff
@@ -46,29 +49,57 @@ func HasConfig(ctx context.Context, pkg *packages.Package, validate bool) (bool,
...
-		tracer.ServiceName("orchestrion-config"),
+	span, ctx := tracer.StartSpanFromContext(ctx, "Load",
+		tracer.ServiceName("github.com/DataDog/orchestrion/internal/injector/config"),
 		tracer.ResourceName(l.dir),
 		tracer.Tag("validate", l.validate),
 	)
 	defer func() { span.Finish(tracer.WithError(err)) }()
 
-	pkgs, err := l.packages(ctx, l.dir)
+	pkgs, err := l.packages(ctx, ".")
```
eliottness, line 102: What does this do ?

.github/workflows/validate.yml
RomainMuller: If they flake, it's easier to re-run just those that flaked to get a complete 6-sample shot for cheaper.

internal/toolexec/proxy/compile.go
RomainMuller: Direct constant concat (compile-time) versus going through the format machinery for it.

internal/jobserver/pkgs/load.go
RomainMuller: These are _already_ shuffled across as JSON internally by the `packages.Load` machinery... 🫣

internal/injector/config/config.go
RomainMuller: I think of it just as a short-hand for the function type... I don't want to force casting non-literal functions to it...

internal/injector/config/config.go
RomainMuller: Loads the package in the current directory (it's canon; and makes it easier to get a cache hit while storing less data)

eliottness: My question was wrong. I was asking, what does this do in here? Does this fix something ? If this `l.dir` value is always equal to `"."` why keep it here ?

.github/workflows/validate.yml
```diff
@@ -271,7 +271,7 @@ jobs:
       - name: Download benchmark reports (raw)
         uses: actions/download-artifact@cc203385981b70ca67e1cc392babf9cc229d5806 # v4
         with:
-          pattern: Benchmark Report (run *)
+          pattern: benchmarks-*.yxy
```
eliottness, line 274:
```suggestion
          pattern: benchmarks-*.txt
```

eliottness (changes requested):

eliottness (approved):

PR #591, opened 2025-04-01
fix: fingerprint mismatch when using go.work files

RomainMuller:
When a `go.mod` file contains a `replace` directive for a module that is part of the current workspace (and directing to the in-workspace directory), the `go list` command does not return a `Replace` field for that module; instead it returns a blank `Version` field.

This commit fixes the file fingerprinting decision so that it hashes the modules' content when the `Version` field is blank or when the module is replaced by a local directory.

Also added a unit test to validate things work as expected in this case.

### DIFF
--- a/internal/jobserver/buildid/version.go
+++ b/internal/jobserver/buildid/version.go
@@ -114,6 +114,18 @@ type moduleInfo struct {
 	Files map[string]struct{}
 }
 
+// shouldHashContent determines whether this module's files need to be hashed in
+// order to properly invalidate build caches when the module changes. This is
+// necessary when the module does not have a version, or is replaced by a module
+// with no version. Both cases imply the module has been replaced by some local
+// directory.
+// Interestingly, the [packages.Module.Replace] field is nil for replaced
+// modules that belong to the current go.work workspace. This is why we also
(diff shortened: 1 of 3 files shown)


rarguelloF (approved):

internal/jobserver/buildid/version.go
```diff
@@ -151,8 +151,13 @@ func (m *moduleInfo) MarshalJSON() ([]byte, error) {
...
 
-	// If this module is replaced by a directory; we'll hash the files as well...
-	if m.Replace != nil && m.Replace.Version == "" {
+	// If this module does not have a version, or is replaced by a module with no
+	// version, it means it has been replaced by some directory on the local
+	// machine (the [packages.Module.Replace] field is nil for replaced modules
+	// that belong to the current go.work workspace). In these cases, we need to
+	// hash the contents of all the files that contribute to the module's build to
+	// properly invalidate when a local file has changed.
+	if m.Version == "" || (m.Replace != nil && m.Replace.Version == "") {
```
kakkoyun, line 160: nit: At this point, maybe it worths having a named predicate function for this logic.

kakkoyun (approved):
LGTM

PR #628, opened 2025-05-15
feat(injector): Support generic join point types

colinodell:
This PR allows Orchestrion's `function` join point to correctly match methods on receivers that are instantiations of generic types (e.g., `*MyType[T]`).

## Problem

Currently, when a method receiver is a generic type like `*MyType[T]`, the Go AST structure is more complex than a simple identifier. For `github.com/dave/dst` nodes, it looks something like this:

```
*dst.StarExpr (represents the '*')
  └── X: *dst.IndexExpr (represents MyType[T])
        ├── X: *dst.Ident (represents MyType - the base generic type)
        └── Index: *dst.Ident (represents T - the type parameter)
```
(For multiple type parameters like `MyType[T, U]`, it would be `*dst.IndexListExpr` with multiple `Indices`).

However, the existing `join.TypeName.Matches` method lacked the necessary cases to handle the `*dst.IndexExpr` or `*dst.IndexListExpr`. It would therefore fail to match any user-specified base type like `MyType` to method receivers like `MyType[T]`, meaning that generic types could not be targeted.

## Solution

I've updated the `join.TypeName.Matches` method to include cases for `*dst.IndexExpr` and `*dst.IndexListExpr`. When these nodes are encountered (representing an instantiated generic type), the matching logic now correctly compares the user-provided type against the base type of the generic (e.g., `MyType`).

### DIFF
--- a/internal/injector/aspect/join/join.go
+++ b/internal/injector/aspect/join/join.go
@@ -108,6 +108,12 @@ func (n TypeName) Matches(node dst.Expr) bool {
 	case *dst.StarExpr:
 		return n.pointer && (&TypeName{path: n.path, name: n.name}).Matches(node.X)
 
+	case *dst.IndexExpr:
+		return !n.pointer && n.Matches(node.X)
+
+	case *dst.IndexListExpr:
+		return !n.pointer && n.Matches(node.X)
+
 	case *dst.InterfaceType:
 		// We only match the empty interface (as "any")
 		if len(node.Methods.List) != 0 {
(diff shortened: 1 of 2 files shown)


kakkoyun (commented):
Looking good @colinodell 

First of all, thanks a lot for the contribution.

Could you please maybe add some tests? Or extend the existing test?

colinodell:
Sure! It seemed like there were no unit tests for `Matches()` so I've implemented some covering both my changes and the existing functionality.  Let me know if you'd like to see any further revisions here!

PR #631, opened 2025-05-27
feat: orchestrion diff

eliottness:


internal/report/report.go
```diff
@@ -0,0 +1,271 @@
...
+					return fmt.Errorf("infer original path for %s: %w", path, err)
+				}
+
+				files = append(files, file)
+				return nil
+			},
+		)
+	}
+
+	// Sort the files by their modified path relative to the work directory.
```
kakkoyun, line 64: 👍

internal/report/report.go
```diff
@@ -0,0 +1,271 @@
...
+		}),
+	}
+}
+
+// Diff generates a diff between the original and modified files and writes it to the writer.
+func (r Report) Diff(writer io.Writer) error {
+	var errs []error
+
+	// Check if the diff command is available, exit early if not
+	if _, err := exec.LookPath("diff"); err != nil {
```
kakkoyun, line 191: Maybe in the subsequent PRs, we can consider `git diff` facilities. e.g. I use fancy git diff setup maybe it is possible to use whatever is set for git.

internal/report/report.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func (r Report) diff(writer io.Writer, file ModifiedFile) error {
+	// If originalPath does not exists, it means that we have cgo files in there, just skip it
+	if _, err := os.Open(file.original); os.IsNotExist(err) {
+		return nil
+	}
+
+	// When adding options, always make sure this is supported on all platforms.
+	cmd := exec.Command("diff",
```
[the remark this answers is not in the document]
eliottness: I tried basically all of them and even combining them but none of them have context window diffs which is a necessity when stdlib files makes multiple thousands lines long

PR #637, opened 2025-06-03
chore: use GitHub App instead of PAT

RomainMuller:


.github/workflows/deps-update.yml
```diff
@@ -13,8 +13,6 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       changes-needed: ${{ steps.is-tree-dirty.outputs.result }}
-    env:
-      GOTOOLCHAIN: local # Prohibits adding `toolchain` directives to go.mod files.
```
eliottness, line 17: Why remove this ?

.github/workflows/pr-labeler.yml
```diff
@@ -19,8 +19,9 @@ jobs:
         with:
           go-version: oldstable
           cache-dependency-path: '**/go.mod'
+
       - name: Assign Labels
         run: go -C _tools run ./conventionalcommit
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ github.token }}
```
eliottness, line 26: How does this work ?

.github/workflows/deps-update.yml
RomainMuller: Noted in the corresponding git commit message; this makes `go get -u` fail when it evaluates whether a given version is usable or not...

The workflow installs `go@<current>` which limits `-u` to only those modules that have a compatible `go` directive; however the version resolver just fails when it sees one such package if `GOTOOLCHAIN=local` instead of simply discarding it... Otherwise, it'll evaluate the package, discover it is not compatible (no shit Sherlock!), and drop it from the candidates list.

.github/workflows/pr-labeler.yml
RomainMuller: Basically, `github.token` is equivalent to `secrets.GITHUB_TOKEN`. The main difference being that workflow validators are able to tell you when that value is not available (as this is documented in the schemas) whereas `secrets` properties have different availability rules.

eliottness (approved):

kakkoyun (approved):

PR #650, opened 2025-06-23
Update ASM Product name to AAP

M4TTRX:
We changed name so I believe that we should reflect that here as well. Let me know if there is a placce where we should mention that asm = aap, as the rename might cause confusion with some documentation.


eliottness:
@M4TTRX Please rebase with signed commits :pray:

PR #667, opened 2025-07-18
CI: Add typo check workflow

atilsensalduz:
This PR adds a typo check to CI using the typos GitHub Action. I also created a typos.yaml config file so we can ignore certain words that are either project specific, technical terms, or intentionally spelled that way. Without this config, the action would flag a lot of false positives, so this should make the checks more useful and less noisy.

If you notice any words in the whitelist that actually are typos and should be fixed, feel free to update the config or let me know. For example, I wasn’t sure whether to change “qualifed” to “qualified”, since it appears in a lot of places and might be intentional.

https://github.com/crate-ci/typos


kakkoyun (approved):
Looking good. 

Thanks for the contribution.

atilsensalduz:
I just noticed that the repository is already using the [typos tool](https://github.com/atilsensalduz/orchestrion/blob/main/.github/workflows/validate.yml#L162-L166), but only for checking documentation. Should I leave spelling workflow as is, or would it be better to add the config file to the existing validate workflow and run it on all files? Alternatively, I can remove the new separate workflow for spelling and stick with checking only the docs as before?
Sorry for the inconvenience. I didn’t notice this before opening the PR. 🤦

kakkoyun:
> I just noticed that the repository is already using the [typos tool](https://github.com/atilsensalduz/orchestrion/blob/main/.github/workflows/validate.yml?rgh-link-date=2025-07-18T09%3A42%3A23Z#L162-L166), but only for checking documentation. Should I leave spelling workflow as is, or would it be better to add the config file to the existing validate workflow and run it on all files? Alternatively, I can remove the new separate workflow for spelling and stick with checking only the docs as before? Sorry for the inconvenience. I didn’t notice this before opening the PR. 🤦

Nothing to worry about. As a reviewer, I have forgotten we have added it.
Feel free to improve the existing setup.

RomainMuller:
<del>@atilsensalduz - this repository can (currently) only accept commits with valid signatures... Would you be able to set up git commit signing (see GitHub's guide @ https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and then update this PR to contain only signed commits?</del>

<del>Otherwise, we'll need to update this repository settings to accept un-signed commits from third-party contributors (it's fine for us to do this but will take a short while).</del>

Never mind, I've just re-committed these leaving you as the author on record.

atilsensalduz:
Really appreciate the merge, @RomainMuller and @kakkoyun  I'll set up GPG signing for my future commits

PR #679, opened 2025-08-12
fix(bench): fix traefik build

rarguelloF:
Traefik fails to build with the following error:

```
webui/embed.go:10:12: pattern static: no matching files found
```

To fix it, before building, create a static folder with some mock content so go embed does not fail.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -107,9 +107,16 @@ func benchmarkGithub(owner string, repo string, build string, testbuild bool) fu
 			// If there's a vendor dir, we need to update the `modules.txt` in there to reflect the replacement.
 			tc.exec(b, "go", "mod", "vendor")
 		}
-		// traefik fails to build if we don't upgrade the version go.opentelemetry.io/otel/sdk/log
+		// traefik needs a few tweaks in order to build successfully
 		if repo == "traefik" {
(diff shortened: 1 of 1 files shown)


main_test.go
```diff
@@ -107,9 +107,16 @@ func benchmarkGithub(owner string, repo string, build string, testbuild bool) fu
...
-		// traefik fails to build if we don't upgrade the version go.opentelemetry.io/otel/sdk/log
+		// traefik needs a few tweaks in order to build successfully
 		if repo == "traefik" {
+			// it fails to build if we don't upgrade the version go.opentelemetry.io/otel/sdk/log
 			tc.exec(b, "go", "get", "go.opentelemetry.io/otel/sdk/log@latest")
+
+			// it fails to build if ./webui/static does not exist, so just create an empty folder
+			if stat, err := os.Stat(filepath.Join(tc.dir, "webui")); err == nil && stat.IsDir() {
+				tc.exec(b, "mkdir", "-p", "./webui/static")
+				tc.exec(b, "touch", "./webui/static/index.html")
```
RomainMuller, line 118: Shelling out to create directories and files? Feels a little excessive?

rarguelloF: changed!

PR #681, opened 2025-08-13
chore: don't re-run all CI on push to main

RomainMuller:
The merge group checks have already checked the same commit(s) that just landed on main, so there is no use in checking these again; it's juts a waste of resources.


kakkoyun (approved):
LGTM

PR #703, opened 2025-09-18
docs(adr): Capture Aspect Application Order decision

kakkoyun:
Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>

PR #702, opened 2025-09-18
chore(docs): Bootstrap Architectural Decisions

kakkoyun:
Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>

PR #704, opened 2025-09-18
feat: Implement advice ordering system

kakkoyun:
Add deterministic and controllable execution order for advice across aspects using namespace
and order fields. 

Includes `OrderableAdvice` interface, sorting logic, and
comprehensive integration tests.

Right now, only "prepend statements" implements `OrderableAdvice`. 
We will implement more when needed.

Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>

### DIFF
--- a/internal/injector/aspect/advice/order_test.go
+++ b/internal/injector/aspect/advice/order_test.go
@@ -35,57 +35,57 @@ func (*nonOrderableAdvice) Apply(context.AdviceContext) (bool, error) { return f
 func (*nonOrderableAdvice) Hash(*fingerprint.Hasher) error            { return nil }
 
 func TestSort_BasicSorting(t *testing.T) {
-	advices := []*OrderedAdvice{
+	advice := []*OrderedAdvice{
 		NewOrderedAdvice("aspect1", &mockOrderableAdvice{id: "z-ns", namespace: "z-namespace", order: 10}, 0),
 		NewOrderedAdvice("aspect2", &mockOrderableAdvice{id: "a-ns", namespace: "a-namespace", order: 10}, 1),
 	}
 
-	Sort(advices)
+	Sort(advice)
 
-	require.Len(t, advices, 2)
-	assert.Equal(t, "a-ns", advices[0].Advice.(*mockOrderableAdvice).id)
-	assert.Equal(t, "z-ns", advices[1].Advice.(*mockOrderableAdvice).id)
+	require.Len(t, advice, 2)
+	assert.Equal(t, "a-ns", advice[0].Advice.(*mockOrderableAdvice).id)
+	assert.Equal(t, "z-ns", advice[1].Advice.(*mockOrderableAdvice).id)
 
-	advices = []*OrderedAdvice{
+	advice = []*OrderedAdvice{
 		NewOrderedAdvice("aspect1", &mockOrderableAdvice{id: "order-20", namespace: "same", order: 20}, 0),
 		NewOrderedAdvice("aspect2", &mockOrderableAdvice{id: "order-10", namespace: "same", order: 10}, 1),
 	}
 
-	Sort(advices)
+	Sort(advice)
 
-	require.Len(t, advices, 2)
-	assert.Equal(t, "order-10", advices[0].Advice.(*mockOrderableAdvice).id)
-	assert.Equal(t, "order-20", advices[1].Advice.(*mockOrderableAdvice).id)
+	require.Len(t, advice, 2)
+	assert.Equal(t, "order-10", advice[0].Advice.(*mockOrderableAdvice).id)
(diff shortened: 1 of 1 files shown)


kakkoyun:
* **#704** <a href="https://app.graphite.dev/github/pr/DataDog/orchestrion/704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/orchestrion/704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#702** <a href="https://app.graphite.dev/github/pr/DataDog/orchestrion/702?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#703](https://github.com/DataDog/orchestrion/pull/703) <a href="https://app.graphite.dev/github/pr/DataDog/orchestrion/703?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

internal/injector/aspect/advice/block.go
```diff
@@ -50,20 +68,44 @@ func (a *prependStatements) Apply(ctx context.AdviceContext) (bool, error) {
...
 }
 
 func (a *prependStatements) AddedImports() []string {
 	return a.Template.AddedImports()
 }
 
+func (a *prependStatements) Order() int {
+	if a.order == 0 {
+		return DefaultOrder
+	}
```
RomainMuller, line 81: No biggie -- found this a little odd, considering `DefaultOrder` is also `0`. Maybe the default should be set at creation time?

internal/injector/aspect/advice/order.go
```diff
@@ -0,0 +1,64 @@
...
+
+package advice
+
+import (
+	"sort"
+)
+
+const (
+	// DefaultNamespace is used when no namespace is specified
+	DefaultNamespace = "default"
```
RomainMuller, line 14: So the namespace participates in ordering -- in the sense that advice is ordered by lexicographic order of `namespace`, then `order`...

And now `default` occupies an arbitrary position that might be...surprising? Using an empty string would, IIUC, make non-namespaced advice consistently apply before namespaced advice... Or maybe we want them to consistently apply after all namespaced advice?

Anyhow -- I think the choice of this value is not neutral and folks _may_ occasionally need to be able to decide if they're before or after this (or do they?). Furthermore, changing this value in the future is a potentially breaking change, so we have to ask ourselves the question _now_.

internal/injector/aspect/advice/order.go
```diff
@@ -0,0 +1,64 @@
...
+		if orderedAdvices[i].namespace != orderedAdvices[j].namespace {
+			return orderedAdvices[i].namespace < orderedAdvices[j].namespace
+		}
+
+		if orderedAdvices[i].order != orderedAdvices[j].order {
+			return orderedAdvices[i].order < orderedAdvices[j].order
+		}
+
+		return orderedAdvices[i].Index < orderedAdvices[j].Index
+	})
```
RomainMuller, line 63: Can we extract the comparison function to a helper function? That'd make it easier to independently test.

internal/injector/aspect/advice/order_test.go
RomainMuller, line 1: Maybe these should include scenarios that involve the `DefaultNamespace`?

internal/injector/config/schema.json
```diff
@@ -881,8 +881,22 @@
...
+              {
+                "properties": {
+                  "order": {
+                    "description": "The execution order within the namespace. Lower values execute first.",
+                    "type": "integer"
+                  },
+                  "namespace": {
+                    "description": "The logical grouping namespace for this advice.",
+                    "type": "string"
+                  }
```
RomainMuller, line 896: So it is legal to specify `order` but not `namespace` (and vice-versa)?

I'm cool with this if intentional... it might be easier to reason over consequences of setting it if one implies the other is required? (Similar to the comment above on the selection of the `DefaultNamespace` value; which BTW should probably be mentioned here for documentation purposes).

.typos.toml
```diff
@@ -11,3 +11,4 @@ Typ = "Typ"
 fo = "fo"
 typ = "typ"
 Unknwon = "Unknwon"
+Advices = "Advices"
```
RomainMuller, line 14: The plural of "advice" is "advice"... Same as "sheep" (which multiple of is "sheep", not "advice" 🥸)

RomainMuller (commented):
Looks pretty good, some tactical comments on certain decisions we have to think hard and deep about before they're impossible to modify...

PR #788, opened 2026-02-13
fix: //orchestrion:ignore is not respected for Echo handlers

### LINKED ISSUE #709: [BUG] `//orchestrion:ignore` directive not respected for Echo router handlers
## Bug Description
The `//orchestrion:ignore` directive is not preventing traces from being emitted for handlers when using Echo router with orchestrion. Despite adding the ignore directive to the readiness handler, traces are still being generated.

## Environment
- **Framework**: Echo router
- **Deployment**: Kubernetes
- **Orchestrion**: latest

### Handler with Ignore Directive
```go
//orchestrion:ignore
func (h *HealthCheckHandlers) ReadinessProbe(c echo.Context) error {
    // Handler implementation
}
```

### Generated Router Code (code from the work directory)
The orchestrion middleware is being applied at the router level:

```go
router :=
//line <generated>:1
    func() *echo.Echo {
        e :=
//line /Users/kaarthik.raogoflink.com/workspace/go/shipping-fee-service/internal/infrastructure/handlers/http/router.go:13
            echo.New()
//line <generated>:1
        e.Use(__orchestrion_echotrace.Middleware())
        return e
    }()
```

## Analysis
It appears that when orchestrion applies middleware at the router level (`e.Use(__orchestrion_echotrace.Middleware())`), it wraps all handlers regardless of individual `//orchestrion:ignore` directives. The middleware-level instrumentation seems to take precedence over function-level ignore directives.

## Possible Solutions
1. The orchestrion middleware should check for `//orchestrion:ignore` directives on individual handlers before applying tracing
2. Provide a way to exclude specific routes/handlers from middleware-level tracing
3. Allow more granular control over which handlers get instrumented when using router-level middleware

## Additional Context
This issue is particularly relevant for health check endpoints (readiness/liveness probes) in Kubernetes environments, where high-frequency health checks can generate excessive trace data that provides little observability value.

Trighap52:
## Why is this change being made?

`//orchestrion:ignore` was not effective for Echo routes because tracing was injected at router level (`echo.New` -> `e.Use(echotrace.Middleware())`), so ignored handlers were still traced.

This caused high-volume endpoints (for example readiness/liveness probes) to emit traces even when explicitly marked with `//orchestrion:ignore`.

## What is the change?

- Added a built-in aspect that rewrites `echo.New()` to `runtime/echoignore.New()`.
- Added `runtime/echoignore` helper package:
  - creates Echo router
  - installs `echotrace.Middleware` with `WithIgnoreRequest(...)`
  - skips tracing when matched route handler is registered as ignored
- Added injector pre-pass to register ignored Echo handlers by injecting:
  - `var _ = echoignore.RegisterIgnoredEchoHandlerFunc(...)`
- Added runtime tests for ignore handler registration and request-ignore matching.
- Added a samples regression case for ignored Echo handlers and updated snapshots.

## How was this tested?

- `go test ./runtime/echoignore`
- `go test ./internal/injector/...`
- `cd samples && go test -run TestSamples .`

Fixes: #709


RomainMuller:
Hey @Trighap52 the actual aspect that is used in live applications is in github.com/DataDog/dd-trace-go, what is in this repository is just an example for integrators.

Trighap52:
Hi @RomainMuller Thanks, that makes sense. To make sure I move this in the right direction: should the fix live entirely in dd-trace-go’s Echo integration, or would you expect Orchestrion core to expose some generic mechanism for integrations to observe //orchestrion:ignore declarations?
My current understanding is that Echo-specific runtime behavior should move out of orchestrion, but I’m not sure whether the “register ignored handler” part should become a generic Orchestrion hook or be implemented fully inside the dd-trace-go aspect/runtime package.

RomainMuller:
IMHO you'd want to do this fully in the dd-trace-go side… I don't think there would be a clean way to signal arbitrary registrations from "usage-agnostic" orchestrion to "usage-specific" tracer integrations. So in this case we'd probably want to introduce a `dd-trace-go` specific pragma/directive that would be used in this way…

Trighap52:
Thanks @RomainMuller that makes sense. I moved this into dd-trace-go as an Echo-specific proposal instead of trying to make Orchestrion handle this generically. I opened a new PR here: https://github.com/DataDog/dd-trace-go/pull/5139

PR #819, opened 2026-04-10
fix(injector): resolve .AST.Fun.Name crash on SelectorExpr in wrap-expression templates

eliottness:
### What does this PR do?

Fixes a build crash (`can't evaluate field Name in type dst.Expr`) that occurs when a `function-call` join-point matches a qualified call like `http.Get(...)` and the advice template accesses `{{ .AST.Fun.Name }}`.

**Root cause:** `proxyCallExpr.Fun()` returns `dst.Expr` (interface). When the underlying node is a `*dst.SelectorExpr` (e.g., for `http.Get`), `proxySelectorExpr` had no `Name` field or method. Go's `text/template` could not resolve `.Name`, crashing with the error above. The same template works fine for bare/dot-imported calls (`Get(...)`) where `Fun` is a `*dst.Ident` with a promoted `Name` field.

**Fix:** Add a `Name() string` method to `proxySelectorExpr` in a hand-written companion file (`dot_ast_extras.go`) that returns `Sel.Name`. This makes `{{ .AST.Fun.Name }}` work transparently for both node types without any changes to dd-trace-go templates — fully backwards compatible.

### Motivation

Customer escalation APMS-19232: any user calling `http.Get(...)`, `http.Head(...)`, `http.Post(...)`, or `http.PostForm(...)` in a function with a `context.Context` parameter fails to build with orchestrion. The dd-trace-go `net/http` client aspect (`Get|Head|Post|PostForm`) triggers the crash.

### Describe how you validated your changes

- Added end-to-end injector regression test `selector-expr-fun-name/` that exercises exactly the failure path: a `function-call` join-point on `net/http.Get` with input code that has a `context.Context` parameter, forcing the template branch that evaluates `{{ .AST.Fun.Name }}` on a `*dst.SelectorExpr`
- All injector tests pass: `go test ./internal/injector/...` (14 packages, 51s)
- `golangci-lint` clean

### Additional Notes

The fix lives in `dot_ast_extras.go`, a hand-written companion to the auto-generated `dot_ast.proxies.go`. This separation keeps generated code untouched while allowing targeted extensions. A comment in the new file guides future maintainers to add similar extensions there.

The alternative (adding a new `FuncName()` template helper) was rejected because it would require dd-trace-go to depend on a minimum orchestrion version, breaking backwards compatibility.

### DIFF
--- a/internal/injector/testdata/injector/selector-expr-fun-name/config.yml
+++ b/internal/injector/testdata/injector/selector-expr-fun-name/config.yml
@@ -1,0 +1,45 @@
+%YAML 1.1
+---
+# Regression test for: https://github.com/DataDog/orchestrion/issues/APMS-19232
+# When a function-call join-point matches a qualified call like http.Get(...),
+# the Fun field is a *dst.SelectorExpr. Templates accessing {{ .AST.Fun.Name }}
+# must work for both *dst.Ident (dot-imported bare call) and *dst.SelectorExpr
+# (qualified call). This test exercises the SelectorExpr path by ensuring the
+# enclosing function has a context.Context parameter (so the {{ if $ctx }} branch
+# is taken and .AST.Fun.Name is evaluated).
+aspects:
+  - join-point:
+      all-of:
+        - not:
+            import-path: net/http
+        - function-call: net/http.Get
+    advice:
+      - wrap-expression:
+          imports:
+            client: github.com/DataDog/dd-trace-go/contrib/net/http/v2/client
+          template: |-
+            {{- $ctx := .Function.ArgumentOfType "context.Context" -}}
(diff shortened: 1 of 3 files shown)


internal/injector/testdata/injector/selector-expr-fun-name/config.yml
```diff
@@ -0,0 +1,45 @@
+%YAML 1.1
+---
+# Regression test for: https://github.com/DataDog/orchestrion/issues/APMS-19232
```
eliottness:
```suggestion
# Regression test for APMS-19232
```

eliottness:
@codex review

RomainMuller (approved):

PR #829, opened 2026-05-05
feat(internal/injector): add method-call join point

rarguelloF:
Add new `method-call` join point that matches calls by the receiver's resolved type. Example:

```yaml                                                                                                                                                                                                                                                                               join-point:                                                                                                                                                                                                                                                                               
  method-call:
    receiver: "go.uber.org/zap.Logger" # fully qualified type name (package import path + type name) without a pointer sigil.                                                                                                                                                                                                                           
    name: Info # the method to match.
    match: any # possible values are: any (default), pointer-only, or value-only.                                                                                                                                                                                                                                                              
```
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Needed for https://github.com/DataDog/dd-trace-go/pull/4729


internal/injector/aspect/join/method_call_test.go
```diff
@@ -0,0 +1,204 @@
...
+			require.NoError(t, err)
+			require.False(t, tn.Pointer, "receiver must not have a pointer sigil")
+
+			m := MethodCall(tn, spec.Name, spec.Match)
+			assert.Equal(t, tt.wantImport, m.Receiver.ImportPath)
+			assert.Equal(t, tt.wantType, m.Receiver.Name)
+			assert.Equal(t, tt.wantMethod, m.Name)
+			assert.Equal(t, tt.wantMatch, m.Match)
+
+			_ = fn
```
darccio: Why is this needed? I see that `fn` isn't called anywhere else.

internal/injector/aspect/join/method_call.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (m *methodCall) ImpliesImported() []string {
+	if path := m.Receiver.ImportPath; path != "" {
+		return []string{path}
+	}
+	return nil
+}
+
+func (m *methodCall) PackageMayMatch(ctx *may.PackageContext) may.MatchType {
```
darccio, line 53: It seems that a unit test for this method is missing.

darccio (changes requested):

internal/injector/aspect/join/method_call_test.go
rarguelloF: nice catch! refactored the test to actually call `fn` and assert the behavior

internal/injector/aspect/join/method_call.go
rarguelloF: updated!

darccio (approved):

PR #856, opened 2026-07-23
feat(join): match functions by receiver presence

RomainMuller:
- allow the function join point's `receiver` option to accept booleans
- interpret `receiver: true` as matching methods with any receiver type
- interpret `receiver: false` as matching non-method functions and function literals
- preserve receiver type strings for matching methods with a specific receiver type
- update the JSON schema and cover parsing and AST filtering behavior


eliottness:
@codex review

kakkoyun (approved):

PR #868, opened 2026-08-03
fix(ci): mint self-mutation token from a workflow pinned to main (2/2)

darccio:
Step 2 of 2. Stacked on [#866](https://github.com/DataDog/orchestrion/pull/866) — **do not merge until #866 lands on `main`.**

## Why this depends on #866
`validate.yml` now calls `.github/workflows/self-mutation.yml` via `uses: DataDog/orchestrion/.github/workflows/self-mutation.yml@main`. That `@main` reference is resolved by GitHub against the actual tip of `main` in this repo — not against this PR's base branch, not against the merge commit being tested. Until #866 merges, `main` doesn't have that file, so **this PR's own CI will fail to load** (the same zero-jobs, instant-failure symptom documented on #866: [run 30830039834](https://github.com/DataDog/orchestrion/actions/runs/30830039834)). That's expected, not a regression — stacking this PR on #866's branch only makes the *diff* clean for review; it doesn't change how `@main` resolves.

Once #866 merges, this PR's base will auto-retarget to `main` and its CI should go green.

## What this does
- `self-mutation` job in `validate.yml` becomes `uses: DataDog/orchestrion/.github/workflows/self-mutation.yml@main` instead of an inline job, moving the token-minting into the trusted reusable workflow added in #866.
- `job_workflow_ref` in `.github/chainguard/self.github.validate.self-mutation.sts.yaml` now targets `self-mutation.yml@refs/heads/main` instead of `validate.yml@refs/heads/main` (the latter being the unsatisfiable pattern from #846 — see #866 for the full history).
- `GitHub Workflow Linters`' SHA-pin allowlist gains `DataDog/orchestrion`, and the `uses:` line gets `# ratchet:exclude`, since this one reference is intentionally a branch ref rather than a SHA.

## Test plan
- [x] `actionlint` clean locally
- [x] YAML parses (chainguard policy + workflow)
- [ ] Once rebased onto `main` (post-#866), confirm `GitHub Workflow Linters` passes and a PR with a generator diff successfully mints the token and pushes (no `permission denied`)

### DIFF
--- a/.github/chainguard/self.github.validate.self-mutation.sts.yaml
+++ b/.github/chainguard/self.github.validate.self-mutation.sts.yaml
@@ -1,10 +1,18 @@
-# Policy for: .github/workflows/validate.yml (self-mutation job) in DataDog/orchestrion
+# Policy for: .github/workflows/self-mutation.yml in DataDog/orchestrion, which is called by the
+# self-mutation job of .github/workflows/validate.yml.
+#
+# `job_workflow_ref` is anchored to `self-mutation.yml@refs/heads/main` rather than to validate.yml.
+# For `pull_request` events GitHub runs workflows from the PR merge ref, so a claim pattern matching
+# validate.yml can only ever match a ref the pull request itself controls -- which would let a PR
+# rewrite the token-minting steps and use this `contents: write` token for anything. Because
+# `job_workflow_ref` resolves to the *called* reusable workflow, pinning the call at `@main` keeps
+# this claim tied to code that has already landed on the default branch.
 issuer: https://token.actions.githubusercontent.com
 subject_pattern: repo:DataDog/orchestrion:pull_request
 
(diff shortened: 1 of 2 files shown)


darccio:
@codex review

darccio:
Superseded by #866, which now contains the complete fix as a single `workflow_run` pipeline instead of a `workflow_call` reusable workflow. The `uses: DataDog/orchestrion/.github/workflows/self-mutation.yml@main` approach this PR wired up can't resolve until `self-mutation.yml` exists on `main`, and when it can't resolve the entire `Tests` run fails to load with zero jobs (see run https://github.com/DataDog/orchestrion/actions/runs/30830932204) -- stripping every check, including the required `Complete`, from any PR affected. Closing this in favor of #866, which has no such failure mode.

PR #870, opened 2026-08-05
fix(injector): return no match instead of an error for function lookups outside a function

rarguelloF:
With an aspect whose template does an argument lookup:

```yaml
aspects:
  - id: http-client-do
    join-point:
      method-call: {receiver: "net/http.Client", name: Do, match: pointer-only}
    advice:
      - wrap-expression:
          imports: {context: context}
          template: |-
            {{- $ctx := .Function.ArgumentOfType "context.Context" -}}
            {{ .AST.Fun }}(
              {{ index .AST.Args 0 }}.WithContext(
                {{- if $ctx -}}{{ $ctx }}{{- else -}}context.Background(){{- end -}}
              ),
            )
```

the `.Function.ArgumentOfType` call fails the build whenever the matched call has no enclosing function, for example a package-level declaration:

```go
package test

...

// no enclosing function, package-level declaration
var resp, err = pkgClient.Do(pkgReq)
```

```
executing "code.Template" at <.Function.ArgumentOfType>: error calling ArgumentOfType:
no function is present in this node chain
```

`dot.Function()` returns `noFunc{}` when nothing in the node chain is a function, and `noFunc` returned an error from its six search methods (`ArgumentOfType`, `ArgumentThatImplements`, `ResultOfType`, `ResultThatImplements`, `LastResultThatImplements`, `FinalResultImplements`), even though the interface documents them as returning an empty result when nothing matches. A template cannot guard against it, because the error happens during template execution.

They now return their zero value and no error, so the template above takes its `{{ else }}` branch at package scope. `Receiver`, `Name`, `Argument` and `Result` still return the error, since having no function really is a failure for those.

### Aspects affected in dd-trace-go

Eight `wrap-expression` aspects use a `.Function` lookup. Six of them can be hit today, because the matched call returns a value and can therefore sit in a package-level declaration:

- `contrib/net/http` — `Get|Head|Post|PostForm`
- `contrib/IBM/sarama` — `NewSyncProducer`, `NewAsyncProducer`
- `contrib/Shopify/sarama` — `NewSyncProducer`, `NewAsyncProducer`
- `contrib/aerospike/aerospike-client-go.v7` — `Client`, added in DataDog/dd-trace-go#5061

Each of these fails to build on v1.11.0 with the error above:

The two zap aspects (`Logger.*` and `SugaredLogger.*`) do the same lookup but cannot be reached, since those log methods return nothing and so cannot appear in a package-level declaration.

Adds an injector testdata case covering both placements, plus unit tests on `noFunc`.

### DIFF
--- a/internal/injector/aspect/advice/code/dot_function.go
+++ b/internal/injector/aspect/advice/code/dot_function.go
@@ -110,32 +110,36 @@ func (noFunc) Argument(int) (string, error) {
 	return "", errNoFunction
 }
 
-func (noFunc) ArgumentOfType(string) (string, error) {
-	return "", errNoFunction
-}
-
-func (noFunc) ArgumentThatImplements(string) (string, error) {
-	return "", errNoFunction
-}
-
 func (noFunc) Result(int) (string, error) {
 	return "", errNoFunction
 }
 
+// The methods below are searches, so outside of a function they report "not found" instead of an
+// error. This lets templates branch on the result, e.g. `{{ if $ctx }}`, and still apply to nodes
+// at package scope.
+
+func (noFunc) ArgumentOfType(string) (string, error) {
+	return "", nil
+}
+
+func (noFunc) ArgumentThatImplements(string) (string, error) {
+	return "", nil
+}
+
 func (noFunc) ResultOfType(string) (string, error) {
-	return "", errNoFunction
+	return "", nil
 }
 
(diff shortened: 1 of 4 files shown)


rarguelloF:
@codex review

darccio:
Benchmarks failing is expected if the branch hasn't been rebased on `main`.

darccio (approved):
LGTM

PR #876, opened 2026-08-12
fix(toolexec): scope implicit test coverage

RomainMuller:
Match nested package loads to the per-binary coverage scope used by `go test` when the user does not provide `-coverpkg`. Cover packages with tests only in their own test variants, while retaining coverage for command-line packages without test files in ordinary loads.

This prevents nested synthetic dependency archives from replacing Go's selected archives with incompatible fingerprints during multi-package coverage runs. Preserve explicit coverage scopes and add unit and end-to-end regression coverage for test and no-test package targets.

### DIFF
--- a/internal/jobserver/pkgs/resolve.go
+++ b/internal/jobserver/pkgs/resolve.go
@@ -257,12 +257,12 @@ func loadResolvedPackages(ctx context.Context, req *ResolveRequest, log zerolog.
 	}, nil
 }
 
-func scopeInferredTestCoverage(buildFlags []string, mode string, packages []string) []string {
+func scopeInferredTestCoverage(buildFlags []string, mode string, pkgs []string) []string {
 	result := withoutCoverageBuildFlags(buildFlags)
-	if len(packages) == 0 {
+	if len(pkgs) == 0 {
 		return result
 	}
-	result = append(result, "-cover", "-coverpkg="+strings.Join(packages, ","))
+	result = append(result, "-cover", "-coverpkg="+strings.Join(pkgs, ","))
 	if mode != "" {
 		result = append(result, "-covermode="+mode)
 	}
--- a/internal/jobserver/pkgs/resolve.go
+++ b/internal/jobserver/pkgs/resolve.go
@@ -157,6 +157,14 @@ func (s *service) resolve(ctx context.Context, req *ResolveRequest) (ResolveResp
 		resp := maps.Clone(ordinary.response)
 		config := ordinary.config
 		config.Env = resolveEnvironment(ctx, req)
+		config.BuildFlags = slices.Clone(config.BuildFlags)
+		if ordinary.testCoverpkgInferred {
+			coveragePackages := slices.Clone(ordinary.testPackagesWithoutTests)
+			if !slices.Contains(coveragePackages, req.TestVariantFor) {
+				coveragePackages = append(coveragePackages, req.TestVariantFor)
+			}
+			config.BuildFlags = scopeInferredTestCoverage(config.BuildFlags, ordinary.testCoverageMode, coveragePackages)
+		}
 		resp, err = mergeTestVariant(ctx, req, ordinary.packages, resp, config)
 		if err != nil {
 			return resolvedPackageSet{}, err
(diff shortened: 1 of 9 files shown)


internal/jobserver/pkgs/resolve.go
```diff
@@ -157,6 +157,14 @@ func (s *service) resolve(ctx context.Context, req *ResolveRequest) (ResolveResp
 		resp := maps.Clone(ordinary.response)
 		config := ordinary.config
 		config.Env = resolveEnvironment(ctx, req)
+		config.BuildFlags = slices.Clone(config.BuildFlags)
+		if ordinary.testCoverpkgInferred {
+			coveragePackages := slices.Clone(ordinary.testPackagesWithoutTests)
+			if !slices.Contains(coveragePackages, req.TestVariantFor) {
+				coveragePackages = append(coveragePackages, req.TestVariantFor)
+			}
+			config.BuildFlags = scopeInferredTestCoverage(config.BuildFlags, ordinary.testCoverageMode, coveragePackages)
```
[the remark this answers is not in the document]
RomainMuller: Fixed in 865826bd. The nested coverage load already contains the synthetic dependency compiled against the covered ordinary target; for external-test-only packages, Go simply does not label that target as a `ForTest` variant. We now retain the transitive importer chain from that load, mark those archives as specific to the test binary, and continue excluding the authoritative package-under-test archive. Added implicit and explicit coverage regressions with a transitive root → middle → subject graph.

eliottness:
@codex review

internal/jobserver/pkgs/resolve.go
```diff
@@ -195,7 +204,16 @@ func loadResolvedPackages(ctx context.Context, req *ResolveRequest, log zerolog.
 		"-a",
 		"-toolexec",
 	)
+	testCoverpkgInferred := goFlags.TestCoverpkgInferred
+	testCoverageMode, _ := goFlags.Get("-covermode")
```
darccio, line 208: When `-covermode` isn't explicit (e.g. relying on `-race` to imply `atomic`), `testCoverageMode` is empty here, so `scopeInferredTestCoverage` omits `-covermode` from the rebuilt scoped flags and lets `go build` re-infer it instead. Since `-race` remains in `buildFlags` (it isn't stripped by `withoutCoverageBuildFlags`), Go should re-infer `atomic` consistently on every scoped rebuild — but can you confirm this holds for every coverage-implying combination, not just `-race` (e.g. `-msan`/`-asan`, or a covermode implied transitively through `GOFLAGS`)? A silent covermode mismatch between the ordinary and test-variant archives would produce incorrect coverage counts rather than a build failure, so it'd fail quietly.

main_test.go
```diff
@@ -105,13 +128,27 @@ func Value() int { return subject.Value() }
 	// Orchestrion injects it into the subject test binary, whose subject archive has coverage.
 	run.exec(t, orchestrion, "go", "test", "-a", "-coverprofile="+filepath.Join(run.dir, "coverage.out"), "./root", "./subject")
 
+	// Without an explicit -coverpkg, Go covers each package that has tests only in its own test
+	// binary, while covering command-line packages without tests in their ordinary form. The nested
+	// load for the subject test binary must therefore leave leaf uninstrumented but instrument
+	// notests, matching the archives against which subject was compiled.
+	run.exec(t, orchestrion, "go", "test", "-a", "-coverprofile="+filepath.Join(run.dir, "coverage-all.out"), "./...")
```
darccio, line 138: This PR's fix hinges on the ordinary and test-variant archives picking the same covermode when it's implied rather than explicit (e.g. via `-race`, per the `resolve.go:208` discussion). No run here exercises that path — every implicit-coverage run uses the default covermode. Worth adding one with `-race` to confirm the two archives stay consistent.

```suggestion
	run.exec(t, orchestrion, "go", "test", "-a", "-coverprofile="+filepath.Join(run.dir, "coverage-all.out"), "./...")

	// Coverage mode implied by -race (rather than an explicit -covermode) must remain consistent
	// between the ordinary archive and each per-binary-scoped test-variant archive.
	run.exec(t, orchestrion, "go", "test", "-a", "-race", "-coverprofile="+filepath.Join(run.dir, "coverage-race.out"), "./...")
```

darccio (changes requested):

internal/jobserver/pkgs/resolve.go
RomainMuller: Covered in 3b2720a4. The scoped-flags unit test now verifies that `-race`, `-msan`, and `-asan` all survive coverage rescoping when no explicit `-covermode` is present. The end-to-end test exercises race-implied atomic mode both directly and through `GOFLAGS=-race`, then checks the resulting profile mode and repeated counter increments. That counter assertion makes a silent set/atomic mismatch fail rather than merely checking that the build succeeds. (Of these sanitizer flags, Go only changes the default covermode for `-race`; `-msan` and `-asan` retain the default set mode.)

main_test.go
RomainMuller: Added in 3b2720a4. There is now an implicit-coverage `-race` run, plus the same scenario with race supplied through `GOFLAGS`. Both verify that the profile uses atomic mode and that three calls to the test-variant function produce a counter of at least three, so the regression catches silent covermode mismatches as well as build failures.

darccio (approved):
LGTM

PR #879, opened 2026-08-18
fix(toolexec): rebuild synthetic test imports

### LINKED ISSUE #875: [BUG] Nested coverage can select incompatible test archives
**Describe what happened:**

`orchestrion go test -cover` can still select archives with incompatible
fingerprints when multiple packages under test participate in a synthetic
link-time dependency chain.

Go's default test coverage mode instruments each package-under-test's test
variant, but not that package's ordinary variant. Orchestrion's nested package
loads currently approximate this behavior with an inferred `-coverpkg` value.
That instruments more package variants than the parent build, including ordinary
variants that the parent test binary built without coverage.

A fingerprint mismatch can therefore occur when a synthetic link-time dependency
imports another package that is also under test and has tests of its own.

Nested package loads should reproduce the coverage configuration of the specific
test binary being linked, so synthetic dependencies resolve to archives with the
same instrumentation and fingerprints as the parent build.

This likely requires configuring coverage per test binary rather than expressing
the parent build's behavior through one inferred `-coverpkg` value.

Add a regression test with two actively tested packages where one package's
synthetic link-time dependency imports the other package and the imported package
has its own tests, then run both packages with coverage enabled.

**Context:**

This limitation was identified and documented while fixing the broader
test-variant coverage failures in
https://github.com/DataDog/orchestrion/pull/872.

RomainMuller:
Implicit test coverage gives the package under test a binary-local archive. Instrumentation-added compiler imports are absent from the Go package graph, so ordinary importer archives can retain a conflicting fingerprint and fail at link time.

Preserve satisfied import provenance and rebuild affected importer closures in a target-specific tool flavor. Carry authoritative archives through nested compilation, record validated closure metadata on the test main, and replay or reconstruct it at link time without polluting ordinary build cache entries.

Keep same-package cycles rejected and add regression coverage for external tests, cache reuse, explicit coverage, and stale metadata.

Fixes #875 (to the extent of what's possible)

### DIFF
--- a/internal/jobserver/pkgs/reversevariants.go
+++ b/internal/jobserver/pkgs/reversevariants.go
@@ -1,0 +1,270 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2023-present Datadog, Inc.
+
+package pkgs
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+
+	"github.com/DataDog/orchestrion/internal/toolexec/archive"
+	"github.com/rs/zerolog"
+	"golang.org/x/tools/go/packages"
+)
+
+const (
+	envVarReverseVariant       = "ORCHESTRION_REVERSE_VARIANT"
+	envVarReverseVariantFlavor = "ORCHESTRION_REVERSE_VARIANT_FLAVOR"
+)
+
+type reverseVariantEnvironment struct {
+	Flavor       string            `json:"flavor"`
+	PackageFiles map[string]string `json:"packageFiles"`
+}
+
+// ReverseVariantEnvironment returns the flavor and authoritative package files
+// for a nested build that reconstructs synthetic importers for a test binary.
+func ReverseVariantEnvironment() (flavor string, packageFiles map[string]string, active bool, err error) {
+	path := os.Getenv(envVarReverseVariant)
+	if path == "" {
+		return "", nil, false, nil
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return "", nil, true, fmt.Errorf("reading reverse test variant environment %q: %w", path, err)
+	}
+	var value reverseVariantEnvironment
+	if err := json.Unmarshal(data, &value); err != nil {
+		return "", nil, true, fmt.Errorf("parsing reverse test variant environment %q: %w", path, err)
+	}
+	if value.Flavor == "" {
+		return "", nil, true, fmt.Errorf("reverse test variant environment %q has no flavor", path)
+	}
+	if flavor := os.Getenv(envVarReverseVariantFlavor); flavor != value.Flavor {
+		return "", nil, true, fmt.Errorf("reverse test variant environment %q has flavor %q, process declares %q", path, value.Flavor, flavor)
+	}
+	if len(value.PackageFiles) == 0 {
+		return "", nil, true, fmt.Errorf("reverse test variant environment %q has no package files", path)
+	}
+	return value.Flavor, value.PackageFiles, true, nil
+}
+
+func (s *service) resolveReverseTestVariant(ctx context.Context, req *ResolveRequest, log zerolog.Logger) (resolvedPackageSet, error) {
+	if req.Pattern == "" || req.TestVariantFor == "" || req.AuthoritativeTarget == "" {
+		return resolvedPackageSet{}, fmt.Errorf("incomplete reverse test variant request for %q", req.Pattern)
+	}
+
+	ordinaryReq := *req
+	ordinaryReq.Pattern = req.TestVariantFor
+	ordinaryReq.TestVariantFor = ""
+	ordinaryReq.ReverseTestVariant = false
+	ordinaryReq.AuthoritativeTarget = ""
+	ordinaryHash, err := ordinaryReq.hash()
+	if err != nil {
+		return resolvedPackageSet{}, err
+	}
+	ordinary, err := s.resolved.Load(ordinaryHash, func() (resolvedPackageSet, error) {
+		return loadResolvedPackages(ctx, &ordinaryReq, log)
+	})
+	if err != nil {
+		return resolvedPackageSet{}, err
+	}
+	if ordinary.buildFlagsErr != "" {
+		return resolvedPackageSet{}, fmt.Errorf("obtaining Go build flags for reverse test variant resolution: %s", ordinary.buildFlagsErr)
+	}
+
+	config := ordinary.config
+	config.Context = ctx
+	config.Env = resolveEnvironment(ctx, req)
+	config.BuildFlags = slices.Clone(config.BuildFlags)
+	if ordinary.testCoverpkgInferred {
+		coveragePackages := slices.Clone(ordinary.testPackagesWithoutTests)
+		if !slices.Contains(coveragePackages, req.TestVariantFor) {
+			coveragePackages = append(coveragePackages, req.TestVariantFor)
+		}
+		config.BuildFlags = scopeInferredTestCoverage(config.BuildFlags, ordinary.testCoverageMode, coveragePackages)
+	}
+
+	return buildReverseTestVariant(req, config)
+}
+
+func buildReverseTestVariant(req *ResolveRequest, config packages.Config) (resolvedPackageSet, error) {
+	// Detect same-package test augmentation before comparing the authoritative
+	// archive. Reconstructing a reverse edge through that augmented package would
+	// require solving a compiler-fingerprint cycle, so preserve the explicit
+	// safety failure instead of reporting a misleading coverage mismatch.
+	provenanceConfig := config
+	provenanceConfig.Tests = true
+	provenanceConfig.Mode = packages.NeedName | packages.NeedImports | packages.NeedDeps | packages.NeedForTest
+	provenance, err := packages.Load(&provenanceConfig, req.TestVariantFor)
+	if err != nil {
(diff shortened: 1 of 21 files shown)


internal/toolexec/archive/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		if hdr.Name != "_go_.o" {
+			continue
+		}
+		data, err := io.ReadAll(io.LimitReader(rd, objectHeaderPrefix))
+		if err != nil {
+			return "", fmt.Errorf("reading Go object %q in %q: %w", hdr.Name, filename, err)
+		}
+		index := bytes.Index(data, []byte(goObjectMagic))
+		if index < 0 || len(data) < index+len(goObjectMagic)+fingerprintSize {
```
darccio: When the magic is found but fewer than 8 trailing bytes remain, this falls
through to the next loop iteration. Since a Go archive has exactly one
`_go_.o` member, that next iteration only reaches `io.EOF` and reports
`ErrNoFingerprint` — correct, but the `continue` reads as if there might be
another `_go_.o` to try. A short comment would save the next reader from
re-deriving that.

```suggestion
		if index < 0 || len(data) < index+len(goObjectMagic)+fingerprintSize {
			// No further "_go_.o" member exists in a Go archive; this falls
			// through to the io.EOF branch above and reports ErrNoFingerprint.
			continue
		}
```

internal/toolexec/archive/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+
+	"github.com/blakesmith/ar"
+)
+
+// ErrNoFingerprint reports an archive without compiler export-data identity.
+var ErrNoFingerprint = errors.New("Go archive has no object fingerprint")
+
+const (
+	goObjectMagic      = "\x00go120ld"
```
darccio: `goObjectMagic` is tied to the Go compiler's object-file format version. If a
future toolchain bumps this magic, `Fingerprint` won't error — it silently
returns `ErrNoFingerprint`, which `CompatibilityFingerprint` then maps to the
literal `"none"` for every archive built with that toolchain. That's a quiet
degradation of the whole compatibility check rather than a loud failure.
Worth a comment recording that this is a known, accepted tradeoff (if it is).

```suggestion
	goObjectMagic      = "\x00go120ld" // Go's export-data object format version; a toolchain bump would silently degrade matches to "none" rather than error.
```

internal/toolexec/archive/fingerprint_test.go
```diff
@@ -0,0 +1,64 @@
...
+	object := []byte("go object header\n\x00go120ld\x01\x23\x45\x67\x89\xab\xcd\xeftrailing data")
+	require.NoError(t, writer.WriteHeader(&ar.Header{Name: "_go_.o", Mode: 0o644, Size: int64(len(object))}))
+	_, err = writer.Write(object)
+	require.NoError(t, err)
+	require.NoError(t, file.Close())
+
+	fingerprint, err := Fingerprint(path)
+	require.NoError(t, err)
+	assert.Equal(t, "0123456789abcdef", fingerprint)
+}
```
darccio, line 41: `TestFingerprint` only covers the happy path (magic present, well-formed).
There's no test for the `ErrNoFingerprint` / `"none"` fallback path — e.g. an
archive whose `_go_.o` has no `\x00go120ld` magic, or one with no `_go_.o`
member at all. That's the exact path `CompatibilityFingerprint` exists to
handle, and it's what two genuinely different no-export archives will both
collapse to in the contract hash (see the flavor-collision question raised
for Area C) — worth locking in with a regression test now.

internal/toolexec/archive/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+		if errors.Is(err, io.EOF) {
+			return "", fmt.Errorf("%w: %q", ErrNoFingerprint, filename)
+		}
+		if err != nil {
+			return "", fmt.Errorf("reading Go archive %q: %w", filename, err)
+		}
+		if hdr.Name != "_go_.o" {
+			continue
+		}
+		data, err := io.ReadAll(io.LimitReader(rd, objectHeaderPrefix))
```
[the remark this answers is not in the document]
RomainMuller: Addressed in `111da8ce`. `Fingerprint` now parses the documented variable-length textual object header through `\n!\n`, bounded to the `_go_.o` archive member, and then validates the object magic and fingerprint. The regression test compiles the same source with short and 5,000-byte build IDs and confirms that both fingerprints match. Malformed or truncated object members now return an error instead of being treated as fingerprint-free.

internal/jobserver/pkgs/reversevariants.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return buildReverseTestVariant(req, config)
+}
+
+func buildReverseTestVariant(req *ResolveRequest, config packages.Config) (resolvedPackageSet, error) {
+	// Detect same-package test augmentation before comparing the authoritative
+	// archive. Reconstructing a reverse edge through that augmented package would
+	// require solving a compiler-fingerprint cycle, so preserve the explicit
+	// safety failure instead of reporting a misleading coverage mismatch.
+	provenanceConfig := config
```
darccio, line 104: This same-package-test check duplicates work already done a moment earlier:
`resolveReverseTestVariants` (oncompile-main.go) only reaches this call after
`resolveTestTargetProvenance()` has already flagged `selected.ForTest ==
testVariantFor` — a signal that conflates "unsafe same-package cycle" with
"safe coverage-caused divergence." This function then re-loads
`req.TestVariantFor` with `Tests=true` from scratch purely to disambiguate
those two cases. Functionally correct (verified both checks agree), but it's
a second `packages.Load` invocation on every reverse-variant reconstruction
attempt. Worth a follow-up to thread the disambiguation through the existing
provenance result instead of reloading, if this path turns out to be hot.

internal/toolexec/proxy/testmain_test.go
```diff
@@ -43,3 +65,83 @@ func TestTestMainMetadataIgnoresImportPathSuffix(t *testing.T) {
 	assert.False(t, ok)
 	assert.Empty(t, target)
```
darccio, line 66: No test exercises the `testMainHeaderV1` read branch in `readTestMain`
(testmain.go:184) — the backward-compatibility path this PR adds for
archives written by a pre-v2 orchestrion binary. Low risk in practice (any
orchestrion binary change already busts the tool-version cache key), but as
written it's untested code on the parsing path for cached/mixed-version
archives.

internal/toolexec/aspect/onlink.go
```diff
@@ -133,3 +159,54 @@ func (w Weaver) OnLink(ctx context.Context, cmd *proxy.LinkCommand) (err error)
 
 	return nil
 }
+
+func refreshTestMainPackageFiles(ctx context.Context, info *proxy.TestMainInfo, reg *importcfg.ImportConfig, workDir string) error {
```
darccio: `refreshTestMainPackageFiles` seems to not have test coverage anywhere in this PR
 — no `onlink_test.go` exists, and nothing in `main_test.go` forces the staleness
path (e.g. by evicting or overwriting a cached `PackageFiles` entry between
the compile that records `TestMainInfo` and the link that consumes it). This
is the reconstruct-and-revalidate fallback: it re-fingerprints cached
archives, rebuilds the whole closure from `ReverseRoots` on a mismatch, and
hard-fails if reconstruction still doesn't match the originally recorded
fingerprints. That's the most novel and highest-risk logic in this change,
with several distinct failure modes (missing root, an import path the
rebuild omits, a re-reconstruction that still mismatches) — and the PR
description explicitly lists "stale metadata" among the regression scenarios
added, but this function is the one path that isn't actually exercised.

A minimal regression test would: build a test binary once so a `TestMainInfo`
with non-empty `PackageFiles`/`ReverseRoots` gets recorded, invalidate one
recorded archive path (e.g. point it at a file that no longer exists, or one
whose content no longer matches the recorded fingerprint), then re-run the
link step and assert reconstruction succeeds and produces matching
fingerprints — plus a second case where the rebuilt archive still doesn't
match, asserting the hard error surfaces instead of silently linking the
wrong archive.

darccio (changes requested):

internal/toolexec/archive/fingerprint_test.go
RomainMuller: This is covered by `TestFingerprintErrors`: its `__.PKGDEF`-only archive has no `_go_.o` member, and the test asserts both `ErrNoFingerprint` and the `CompatibilityFingerprint == "none"` fallback. `TestFingerprintRejectsMalformedGoObject` separately covers `_go_.o` members without valid magic; those are deliberately rejected as malformed rather than treated as fingerprint-free.

internal/jobserver/pkgs/reversevariants.go
RomainMuller: Agreed as a follow-up if profiling shows this path is hot. I am leaving it unchanged here: the load is also the only provenance source during link-time reconstruction, where there is no earlier caller-side provenance result to thread through, and reverse requests are memoized by request hash.

internal/toolexec/proxy/testmain_test.go
RomainMuller: This branch is covered by `TestTestMainMetadataV1`, which reads a V1 archive through `LinkCommand.TestMainInfo` and asserts the complete V1-derived `TestMainInfo`. The malformed-metadata table also exercises V1 missing-target and trailing-data cases.

internal/toolexec/aspect/onlink.go
RomainMuller: Addressed in `85c012ae`. I injected the reverse-package resolver into `refreshTestMainPackageFiles` and added regression tests for successful stale reconstruction and hard fingerprint mismatch. The tests also cover resolver failure, omitted imports, and unreadable rebuilt archives, and assert that failed reconstruction does not mutate the cached package-file selection.

darccio (approved):

Issue #687, opened 2025-08-25
[BUG] `orchestrion pin` on clean repo breaks in `pruneImports` with vendoring


RomainMuller:
**Version of orchestrion**
`1.5.0`

**Describe what happened:**
A Dockerfile with these characteristics:

```
# ...

ARG CGO
ARG OS
ARG ARCH
RUN mkdir /go/src/project
WORKDIR /go/src/project
COPY . ./
RUN go install github.com/DataDog/orchestrion@latest
RUN orchestrion pin
RUN go env
RUN echo "CGO = "$CGO
RUN echo "OS = "$OS
RUN echo "ARCH = "$ARCH
RUN CGO_ENABLED=$CGO GOOS=$OS GOARCH=$ARCH govvv build -mod vendor -toolexec='orchestrion toolexec' -a -installsuffix cgo -ldflags '-extldflags "-static"' -o ./bin/project ./cmd/project/project.go
RUN cp ./bin/project /bin/
RUN CGO_ENABLED=$CGO GOOS=$OS GOARCH=$ARCH govvv build -mod vendor -toolexec='orchestrion toolexec' -a -installsuffix cgo -ldflags '-extldflags "-static"' -o ./bin/projectcli ./cmd/projectcli/projectcli.go
RUN cp ./bin/projectcli /bin/
```

Results in the following error:
```
ERROR [project builder 6/14] RUN orchestrion pin 0.3s
------
> [project builder 6/14] RUN orchestrion pin:
0.258 pruning imports from "/go/src/project/orchestrion.tool.go": pruneImports: err: exit status 1: stderr: go: inconsistent vendoring in /go/src/project:
0.258 github.com/DataDog/dd-trace-go/orchestrion/all/v2@v2.2.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
0.258 github.com/DataDog/orchestrion@v1.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
[...]
0.258 To ignore the vendor directory, use -mod=readonly or -mod=mod.
0.258 To sync the vendor directory, run:
0.258 go mod vendor
0.258
------
failed to solve: process "/bin/sh -c orchestrion pin" did not complete successfully: exit code: 255
make: *** [Makefile: 14 :  build] Error 17
```

**Describe what you expected:**

Expected `go mod vendor` to be automatically run at the right time in the process so this error does not occur.

PR #885, opened 2026-09-01
fix(pin): resolve dependency imports from the consuming module, not their own directory

### LINKED ISSUE #760: 📣 Information about "Removing unnecessary import of github.com/DataDog/dd-trace-go/orchestrion/all/v2" warning
If you see this line in your Orchestrion output, it's not an error:

```
Removing unnecessary import of "github.com/DataDog/dd-trace-go/orchestrion/all/v2": in "github.com/DataDog/dd-trace-go/contrib/99designs/gqlgen/v2" imported from "/Users/faraz.oman/gowork/pkg/mod/github.com/!data!dog/dd-trace-go/orchestrion/all/v2@v2.4.0/orchestrion.tool.go": in "github.com/DataDog/dd-trace-go/contrib/99designs/gqlgen/v2": -: github.com/DataDog/dd-trace-go/contrib/99designs/gqlgen/v2@v2.4.0: replacement directory ../../contrib/99designs/gqlgen does not exist
```

We'll work on making sure it doesn't appear, as it's a bug but it doesn't have impact because Orchestrion reintroduces the missing package.

darccio:
## Summary

Fixes #760.

`orchestrion pin` was resolving transitive imports found in a dependency's `orchestrion.tool.go` file (e.g. `github.com/DataDog/dd-trace-go/orchestrion/all/v2`) against **that dependency's own module-cache directory**, instead of the module actually consuming it. This made the dependency's own checkout-relative `replace` directives (valid only inside a `dd-trace-go` checkout, e.g. `=> ../../contrib/...`) get applied during resolution, causing `go list` to fail and the import to be misreported as "unnecessary" and pruned — even though the import was perfectly valid.

```
removing unnecessary import of "github.com/DataDog/dd-trace-go/orchestrion/all/v2": in
"github.com/DataDog/dd-trace-go/contrib/99designs/gqlgen/v2" imported from
".../orchestrion/all/v2@v2.4.0/orchestrion.tool.go": ...: replacement directory
../../contrib/99designs/gqlgen does not exist
```

- `config.HasConfig` now takes an explicit resolution directory (the consuming module's root), separate from the directory a config file is read from.
- `pin`'s package loader now forces `-toolexec=` (matching the jobserver's loader) so resolution never recurses into `orchestrion` itself if `GOFLAGS=-toolexec=...` is set.
- Closed a narrow gap where `go mod tidy` wouldn't run before pruning if `ensure.RequiredIntegrations` produced no edits, which could leave `go.sum` incomplete for the transitive contribs pruning needs to resolve.
- **Guardrail:** pruning now only removes an import once `HasConfig` positively established it provides no integrations (`err == nil && !hasConfig`). Any resolution failure (an unreachable proxy, a broken `replace`, etc.) is reported as a non-fatal `note:` warning on stderr and the import is left untouched — it is only fatal under `-validate`.

## Test plan

- [x] `go test ./internal/injector/config/... ./internal/pin/...` — new hermetic regression tests reproduce the #760 shape both at the `config.HasConfig` level and end-to-end through `PinOrchestrion` (`internal/pin/testdata/relative-replace/`).
- [x] `go build ./...`, `go vet ./...`, `golangci-lint run ./...` all clean.
- [x] Manually reproduced the bug and confirmed the fix with a built binary in a scratch module (no false-positive prune, `orchestrion.tool.go` retains `dd-trace-go/orchestrion/all/v2`, exit 0).

## Out of scope

Flagged as follow-up work, not included here:
- #687 (`orchestrion pin` fails with "inconsistent vendoring") — same function area, different root cause; currently being investigated separately.
- Pruning currently never persists to disk (`writeUpdated` runs before `pruneImports`, so removed imports never actually leave `orchestrion.tool.go` — only `go mod tidy`'s cleanup of unused `go.mod` requires is visible). Fixing this safely depends on #687 landing first, since `go mod vendor` doesn't copy `orchestrion.yml` files, and persisting prunes without vendor-awareness would silently delete valid integrations from vendored projects.
- The no-edits fallback path (`pin.go`'s `if !tidied { gomod.Run(ctx, "tidy", ...) }`) runs `go mod tidy` directly, bypassing the `go mod vendor` step that `gomod.RunEdit` performs when a `vendor/` directory is present. In a vendored project that hits this path (no `go.mod` edits needed), `go.sum` is refreshed but `vendor/` is not — the same vendor-awareness gap as #687, tracked there rather than fixed here.
- The new prune-output wording (`"removed %q from %s: %s"`) is phrased in the past tense, which overstates what happens today given the persistence gap noted above — the import is not actually removed from `orchestrion.tool.go` on disk yet. The wording should be revisited once persistence lands (tracked with #687); not fixed in this PR.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -54,7 +54,7 @@ jobs:
         if: github.event_name != 'push'
         run: |-
           currentTag=$(go run . version -static | cut -d' ' -f 2)
-          find . -iname go.mod -not -path './_docs/themes/**' -print0 | while IFS= read -r -d '' gomod; do
+          find . -iname go.mod -not -path './_docs/themes/**' -not -path '*/testdata/*' -print0 | while IFS= read -r -d '' gomod; do
             dir="$(dirname "${gomod}")"
             if [[ -n $(go -C="${dir}" mod edit --json | jq '(.Replace // []).[] | select(.Old.Path == "github.com/DataDog/orchestrion")') ]]; then
               echo "Aligning orchestrion version in ${dir} to ${currentTag} if necessary..."
--- a/internal/injector/config/config-go.go
+++ b/internal/injector/config/config-go.go
@@ -26,6 +26,33 @@ const FilenameOrchestrionToolGo = "orchestrion.tool.go"
 
 var ErrInvalidGoPackage = errors.New("no .go files in package")
 
+// isNoGoFilesError reports whether e is the (poorly-typed) error emitted by
+// [packages.Load] when a package's directory has no Go files at all -- as
+// opposed to the package failing to resolve in the first place.
+func isNoGoFilesError(e packages.Error) bool {
+	return e.Kind == packages.ListError && strings.Contains(e.Msg, "no Go files in")
+}
+
+// unresolvedError returns a non-nil error if pkg.Errors indicates the package
+// could not be resolved at all (module not found, a `replace` directive
+// pointing at a missing directory, an unreachable proxy, inconsistent
+// vendoring, ...), as opposed to resolving to an empty, Go-file-less
+// directory. A resolution failure must never be mistaken for evidence that the
+// package lacks configuration.
+func unresolvedError(pkg *packages.Package) error {
+	var errs []error
+	for _, e := range pkg.Errors {
+		if isNoGoFilesError(e) {
+			continue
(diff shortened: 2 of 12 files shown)


darccio:
@codex review

kakkoyun (approved):

PR #890, opened 2026-09-01
fix(pin): guard against nil import spec in pruneImports

darccio:
## Summary
- `importSet.Find(pkg.PkgPath)` in `pruneImports` was dereferenced without a nil check; if `go/packages` ever returned a `PkgPath` not exactly matching a registered import (not currently known to be reachable, but not provably impossible), this would panic.
- Add a nil check that logs a warning and skips the decoration step instead of crashing.

Pre-existing code path, not introduced by a recent change — spotted during review of #888 as worth hardening defensively.

## Test plan
- [x] `go build ./internal/pin/...`
- [x] `go vet ./internal/pin/...`
- [ ] No automated regression test added — reliably forcing `packages.Load` to return a mismatched `PkgPath` isn't practical without a larger refactor or network-dependent module tricks, which is out of scope for this minimal hardening fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


darccio:
@codex review

kakkoyun (approved):

kakkoyun:
@copilot resolve the merge conflicts in this pull request

PR #897, opened 2026-09-14
fix(jobserver): reject cyclic never-build-twice waits

eliottness:
## Motivation

dd-trace-go integration jobs exposed a compile loop that caused Orchestrion builds to hang indefinitely. When dependency resolution spawned a nested build of the package currently being compiled, both compile requests had the same import path and build ID. The never-build-twice service treated the nested request as ordinary concurrent work and waited for the original task, while the original task was waiting for the nested build.

## Changes

- Propagate the parent compilation import path on never-build-twice start requests.
- Reject a request when an unfinished compilation would wait on itself, returning a `cycle detected` error instead of blocking.
- Add service-level and end-to-end regressions for the compile-loop scenario.

The runtime fix is limited to a three-line guard in `internal/jobserver/nbt/nbt.go`. Normal duplicate-build reuse, including completed results, is unchanged.

## Testing

- `go test -race -shuffle=on -count=1 ./internal/jobserver/...`
- `go test -tags=e2e -run TestCompileLoop -count=1 ./test/e2e/`
- Manual fixture build: exits in about five seconds with `cycle detected: example.com/compileloop/victim -> example.com/compileloop/victim` instead of hanging.

---

PR by Bits - [View session in Datadog](https://app.datadoghq.com/code/625d6db7-bbe9-4e2f-8462-becf144aabc9)

Comment @datadog to request changes


internal/jobserver/nbt/nbt_test.go
```diff
@@ -380,4 +381,45 @@ func Test(t *testing.T) {
...
+
+	// Compilation tasks resolve the archives of the dependencies injected code
+	// introduced by spawning nested builds, which may have to compile a package one
+	// of their ancestors is already compiling. The artifacts of that task can only
+	// be produced once the nested build completed, so waiting for them deadlocks
+	// the whole build. See TestCompileLoop in `test/e2e` for the same scenario in a
+	// real build.
+	//
+	// KNOWN GAP: this test currently fails, as the job server waits for artifacts
+	// that can never be produced.
```
RomainMuller, line 393: Is this accurate?

RomainMuller (approved):

internal/jobserver/nbt/nbt_test.go
eliottness: no, removing

PR #898, opened 2026-09-14
fix(toolexec): forward unknown tool flags instead of failing

darccio:
## What

The flag sets used by the `-toolexec` proxy are **generated** from the usage output of a specific Go release (`internal/toolexec/proxy/{compile,link}.flags.go`, produced by `internal/toolexec/proxy/generator`). When a newer Go toolchain passes a flag that did not exist in that release, `flag.FlagSet.Parse` fails and every instrumented compilation breaks.

Concrete example seen in the wild: Go 1.28 tip's `cmd/go` now passes a new internal `-exportfd` flag to `compile` (the pipelined "early export data" work, [golang/go#15734](https://github.com/golang/go/issues/15734), CLs 827844 / 828986–828989). Orchestrion's generated flag sets — v1.9.0 against `go1.25`, and even the latest v1.13.0 against `go1.26` — do not know it, so `DataDog/dd-trace-go`'s `Gotip Testing` nightly fails on every package with:

```
# internal/goarch
flag provided but not defined: -exportfd
Usage of compile version go1.25:
  ...
FAIL <pkg> [build failed]
```

Regenerating the flag sets against tip cannot keep up: new compiler flags will keep appearing between orchestrion releases.

## How

Generated `parse()` methods now delegate to a new shared `parseToolFlags` helper that:

- Pre-scans the raw arguments using `flag`-package semantics (flags stop at the first positional or `--`).
- Strips tokens for flags **not** registered in the generated flag set, then parses the remainder normally. Because the proxy keeps the **original argv** and `RunCommand` re-executes it verbatim, stripped unknown flags are automatically forwarded to the proxied tool — no re-insertion logic needed.
- Keeps parsing known flags exactly as before, so typed access to `-importcfg`, `-o`, `-p`, `-trimpath`, `-buildid`, `-buildmode`, `-V`, … is unchanged.

For the ambiguous two-token unknown-flag form (`-unknown value`), the value token is consumed **unless** it is absent, looks like another flag (leading `-`), or looks like a source/object file (`.go`, `.a`, `.o`) — in which case the flag is treated as value-less and the next token stays among positionals. The invariant: **never swallow an input file**, since that would corrupt `CompileCommand.Files` / `LinkCommand.Inputs`-driven logic. A stray token among positionals is harmless (`GoFiles()` filters non-`.go` entries) and never affects the forwarded command line.

The generator now emits the `parseToolFlags` call, so tolerance survives future regenerations. `compile.flags.go` / `link.flags.go` were regenerated with the current stable toolchain (flag-set version tag `go1.26` → `go1.27`, picking up go1.27's `-u` compile flag).

## Testing

- `go build ./...` and `go vet ./internal/toolexec/...` pass.
- `go test ./internal/toolexec/...` passes; new regression cases in `compile_test.go` / `link_test.go` cover `-exportfd=3`, `-exportfd 3`, and a value-less unknown flag before input files — each asserting that known flags still parse **and** `cmd.Args()` equals the original invocation (verbatim forwarding).
- New `flags_test.go` covers helper edge cases (unknown flag after positional, `=` form, two-token form, value-looking-like-flag).
- Sanity-checked end-to-end: a compile invocation containing `-exportfd 3` no longer fails at the proxy parse stage.

## Notes for reviewers

- The two-token heuristic is deliberately conservative; failure modes only ever affect metadata lists (`Files` / `Inputs`), never the forwarded command line. Happy to make it stricter or pluggable if preferred.
- Asm/other tools pass through as `CommandTypeOther` and were already forwarded untouched.
- Scope limitation: this handles flags *unknown* to the generated set (new flags). It does not handle a *known* flag changing shape between Go releases (e.g. a currently-boolean flag becoming value-taking, or vice versa) — the generated flag set would still mis-parse that case. Regenerating against each new toolchain (as usual) is what re-derives the correct shape for known flags; this tolerance is a bridge between regenerations, not a replacement for them.


kakkoyun (approved):

darccio:
Closed because it was opened from my fork.

PR #902, opened 2026-09-15
fix(pin): sync the enclosing go.work with the module's go directive

kakkoyun:
### Summary
- `orchestrion pin` runs `go mod tidy` on the target module, which can raise the module's `go` directive when a newly pinned dependency requires a newer language version (for example, an orchestrion release whose dependency set requires `go 1.26.0`).
- If the module belongs to a workspace, the enclosing `go.work` was left with its older `go` directive, so every subsequent `go` command in that workspace failed with `go: module . listed in go.work file requires go >= 1.26.0, but go.work lists go 1.26`.
- `PinOrchestrion` now calls `gomod.AlignWorkGoVersion` after its tidies: it bumps the `go` directive of the enclosing `go.work` via `go work edit` whenever the module's directive became newer. etcd-style monorepos hit this first, as their generated `go.work` files use the two-part form ("1.26"), which does not satisfy a module's three-part requirement ("1.26.0").

Surfaced by the v1.14.0-rc.1 benchmarks against etcd v3.7.1, but the failure mode does not depend on the release.

### Test plan
- [x] `TestAlignWorkGoVersion` covers the raise, the no-op, and the no-workspace cases.
- [ ] Benchmarks of the v1.14.0-rc.1 release candidate (#901) with this change included.


hannahkm (approved):
lgtm!

PR #903, opened 2026-09-15
release: v1.13.1

darccio:
This is required to avoid issues with Go 1.28 (`gotip`).


darccio:
@kakkoyun We need to release first this patch version, so you'll need to rebase your release #901 on this when it merges into `main`.

hannahkm (approved):

PR #905, opened 2026-09-17
fix(release): authenticate submodule tag publication

kakkoyun:
## Why
Publishing v1.13.1 created the root tag, but the follow-up job could not create instrument/v1.13.1 because organization tag protection rejects the default GITHUB_TOKEN. The missing tag was repaired manually.

## Changes
- Use repository-scoped, short-lived dd-octo-sts credentials, with separate policies for published releases and manual recovery from main.
- Resolve the release commit from the published root tag rather than target_commitish.
- Preserve correct existing tags, reject conflicting tags, and reconcile uncertain API responses.
- Add a manual recovery trigger, regression tests, and recovery documentation.

No protection rules, existing tags, or RC targets are changed.

## Verification
- 15 automation and trust-policy tests passed against local Git repositories.
- go build ./... passed.
- Bash syntax, ShellCheck, shfmt, actionlint, yamlfmt, ratchet, whitespace, and active LSP checks passed.
- Full Go tests were not rerun because no Go source changed. Markdown lint was unavailable locally.

## After approval and merge
Run the recovery workflow from main for v1.13.1 to verify live STS token exchange and the no-change recovery path. Existing tags should remain unchanged. This does not prove permission to create a new protected tag; that still needs live verification.

Already-cut releases retain their original workflow revision and can use manual recovery after publication.


kakkoyun:
@codex review

